PR #64, opened 2017-07-12
Support for azure web app's load balancer

pushrbx:
I was testing squidex under Azure Web App on Linux and I've found that the OIDC doesn't work at all. So I've added some changes which makes it work, however I'm not sure about any other load balancer setups. The current "UseMyForwardingRules" method also can cause problems, becuase if the load balancer communicates with the web server through http instead of https the "EnforceHttpsMiddleware" will cause an endless redirect loop. Also in my opinion the enforment of https is the host's deal, it should be configured in the web server.

### DIFF
--- a/src/Squidex/Config/Web/WebUsages.cs
+++ b/src/Squidex/Config/Web/WebUsages.cs
@@ -36,6 +36,32 @@ namespace Squidex.Config.Web
             app.UseMiddleware<EnforceHttpsMiddleware>();
         }
 
+        public static void UseMyAzureLoadBalancerForwardingRules(this IApplicationBuilder app)
+        {
+            var forwardingOptions = new ForwardedHeadersOptions
+            {
+                ForwardedHeaders = ForwardedHeaders.XForwardedProto | ForwardedHeaders.XForwardedFor,
+                ForwardLimit = null,
+                RequireHeaderSymmetry = false
+            };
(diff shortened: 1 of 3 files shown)


SebastianStehle:
Thanks a lot ...

BUT ... ;)

I am pretty sure that Azure LB supports the X-Forwarded-Proto header. It seems to be standard now. I tested it with Nginx, IIS, Cloudflare and some years ago with Azure LB in other project.

If we want to support more headers it would be better to name these headers in the configuration instead of making a boolean for each Loadbalancer.

pushrbx:
You are right. Also that's right that Azure LB supports the X-Forwarded-Proto header, but not in the latest one, where the app service runs on linux. I'll refactor this to a flexible header config thing.

SebastianStehle:
Strange. Why does it depend on the operation system? It is just an http header.

pushrbx:
Because they are using a different software stack for the managed vms. Azure App Service on Linux uses docker. I guess in the docker image an apache runs, and the output of the dotnet core stuff gets pipelined through apache to a load balancer outside of the container. But this is just a rough guess. The headers are being modified someway during their way from the docker container to the load balancer.

pushrbx:
Invalid. Will be fixed by microsoft anyway.

SebastianStehle:
That is good :) ... thank you very much

Issue #44, opened 2017-05-11
Does it work with Azure Doc/CosmosDb?


ctolkien:
It has a MongoDb compatible API, so it _might_ work? Anyone given it a crack?

SebastianStehle:
I have not tried it yet. I run Mongodb in a Kubernetes cluster at the moment. But it should work. I hope that DocumentDB/CosmosDB supports all driver features.

pushrbx:
I tested it with Cosmos DB, but it doesn't work, because the aggregate queries aren't supported yet. It's still in private preview.
https://stackoverflow.com/questions/44844678/azure-cosmosdb-using-mongo-drivers-get-count-with-out-getting-all-documents-bas

SebastianStehle:
Awesome. It would be nice to get it working, to reduce the maintenance costs.

There are three "databases" at the moment:

1. EventStore
2. Identity (Users)
3. Read Model

1. EventStore
The event store is easy to implement. And because

pushrbx:
There is no need to change the query system to something different, it can stay as is, it is only required to replace those queries which try to do an aggregation query to ones where the system loads the data to memory and does the aggregation on it there, because Cosmos will support it anyway eventually. I'm still trying to wrap my head around the code, but I think it's going to be a different implementation for one of the mongodb related interfaces and a config item to say "we are using cosmos".

SebastianStehle:
Would be awesome :)

pushrbx:
I made some fixes, now I can create users, list the users, modify my profile. (pull request soon) But I can't create an app, because during publishing the event it throws a `MongoDB.Driver.MongoCommandException: Command failed. ` exception. How does the app creation work? What are the steps until the app gets created?

SebastianStehle:
It is an CQRS architecture. Everything is written to the EventStore and then the events are handled to populate read models. I think the EventStore could be the critical part.

SebastianStehle:
I think the problem was the BsonTimestamp, that is not compatible with MongoDB. I am working on an implementation that is not dependent on this data type.

pushrbx:
I could catch the exception finally.
For this query:
`{{ "find" : "Events", "filter" : { "Timestamp" : { "$gte" : Timestamp(0, 0) }, "EventStream" : /^asset-/ }, "sort" : { "Timestamp" : 1 } }}`
CosmosDB responds through the mongodb driver with this:
[...]

PR #65, opened 2017-07-12
Fixes for Cosmos Db compatibility

pushrbx:
Changed some of the aggregate queries to query in to memory then do the aggregation operation on the collection of items. I'll commit more to this pull request as I find more not working queries. Don't merge it yet please. 
Tries to fix issue #44


SebastianStehle:
It would be much easier to assume that there will not be more than 2 Billion users and just use a Count() instead ;)

Retrieving all users just to get the count is very expensive ;)

pushrbx:
I'll try to come up with another solution this weekend.  :)

SebastianStehle:
I have just tested Any() and Count() ... works fine, but the problem is the EventStore :(

SebastianStehle:
I close this pull request for now, because I need to find a solution for the EventStore first. Perhaps just sql? ;)

PR #76, opened 2017-07-19
[AssetStore] Azure Storage support

pushrbx:
Work in progress. Basic features are working. (uploading, downloading assets)

Things to consider:
- this implementation (first commit) is not performance friendly. There are many round trips between azure and the app.

Missing things/things to do:
- Cropping - currently the system returns 404 status for cropped images
- Mime type checking to store the file with the right metadata
- unit tests


SebastianStehle:
Very nice,

I am not sure, what to do with the pull request yet. I would like to close the v1.0 in the master branch and then think about a branching strategy. 

I also have some feedback:

* I use ConfigurationException for what you do with the AzureStorageException at the moment.
* I think it would be better to to make a Squidex.Infrastructure.Azure assembly, because there are other Services from Azure we could use, e.g. ServiceBus.
* I have not used Snapshots yet for such a scenario. I thought there are only useful for Disks. Would it not be easier to use the Id as prefix, e.g. {id}/{version}_{suffix}.blob or so?. Afaik you can also delete by prefix and it would make the code much more easier. The Google Cloud integration has 80 LOC ;)
* What is the BlobContainerProvider and the StorageAccountManager used for? It looks a little bit like overengineering to me ;)

For me there are 2 options right now:

1. Integrate the pull request to a separate branch and make you a contributor
2. Finalize the azure integration in your personal branch.

pushrbx:
My intention was with this pull request with its "first commits" to get a code review from you, so I can improve it. 
This one can go to a later version, because it's not finished yet.

> What is the BlobContainerProvider and the StorageAccountManager used for? It looks a little bit like overengineering to me ;)

I might have tried to seperate the responsibilities too much.

> I have not used Snapshots yet for such a scenario. I thought there are only useful for Disks. Would it not be easier to use the Id as prefix, e.g. {id}/{version}_{suffix}.blob or so?. Afaik you can also delete by prefix and it would make the code much more easier. The Google Cloud integration has 80 LOC ;)

Ok, I will change it. It will make it quicker as well.

> I think it would be better to to make a Squidex.Infrastructure.Azure assembly, because there are other Services from Azure we could use, e.g. ServiceBus.

I agree.

SebastianStehle:
I will change the namespace structure a little bit this afternoon and then merge your code in and test it. :)

pushrbx:
What about the cropping of images? They are not working with this, and I'm not sure how to go about it.

SebastianStehle:
But why? Strange.

pushrbx:
http://localhost:5000/api/assets/763ea844-79e8-4bcd-a756-da54bd9a3b2f?version=0&width=184&height=123&mode=Crop
With azure, this request returns 404. When I remove the query string, it works. So the problem must be around the cropping logic.

SebastianStehle:
Hi, I improved the interface for the Asset store. Furthermore I created a test base class for all stores.

pushrbx:
Tests has been added. Also they are passing. The cropping still needs testing.

SebastianStehle:
Very nice. Will integrate it tomorrow :)

PR #78, opened 2017-07-19
(fix): wrong return type in interceptor

DavidRouyer:
makeRequest expects a return type of Observable<HttpEvent<any>>


SebastianStehle:
You are right, I copied the code from my "Pre-Interceptor" code.

PR #111, opened 2017-09-14
Migration of profile controller to asp.net core 2.0

pushrbx:
@SebastianStehle  I think you have missed this one. It didn't compile for me, so I guess this is the right change based on this: https://docs.microsoft.com/en-us/aspnet/core/migration/1x-to-2x/identity-2x
Tell me if im wrong. :smile:

### DIFF
--- a/src/Squidex/Controllers/UI/Profile/ProfileController.cs
+++ b/src/Squidex/Controllers/UI/Profile/ProfileController.cs
@@ -34,19 +34,16 @@ namespace Squidex.Controllers.UI.Profile
         private readonly IUserPictureStore userPictureStore;
         private readonly IAssetThumbnailGenerator assetThumbnailGenerator;
         private readonly IOptions<MyIdentityOptions> identityOptions;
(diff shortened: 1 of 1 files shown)


SebastianStehle:
I am confused, because I have not migrated to asp.net 2.0 yet. Only to .netcore2.0. The reason is that IdentityServer4 for 2.0 is not stable yet.

pushrbx:
Somehow nuget started complaining for me that this and that package is missing or not compatible with .net core 2.0, and I couldn't build the solution, so I got fed up, and I hit the update all package button.

SebastianStehle:
Really? I fixed all warnings...But had some strange errors too. They published the RC1 10 days ago. So I guess that I can update it very soon. My impression is that I spend 50% of my time with upgrading packages. Either .NET or Javascript/Angular :D

pushrbx:
Gotcha. I did something wrong. Now builds fine. Sorry for the trouble. :smile:

SebastianStehle:
I have another pull request for a full migration.

PR #138, opened 2017-10-11
Complex queries via plugins

pushrbx:
With the help of Autofac assembly scan squidex loads assemblies from a folder and it looks for types which implements `IQueryModule` interface and registers them in the container.
There is an extra controller: `ComplexQueriesController` which looks for the `query` instances from the loaded assemblies based on their reported name and which schema they belong to.
Example request: 
`http://localhost:5000/api/content/portal/node/queries/getNodesWithTag?tagNames=1998,1999,2000`

These `query` objects will be shown in swagger and they are accessible from graphql.

**Note**: This PR still needs a review!
**TODO**: Documentation


SebastianStehle:
I have not reviewed everything yet, but I would change the plugin system.

What I would do:

1. Scan for Modules in all assemblies and load them.
2. Rename IQueryModulesService to IQueryProvider or so

The advantage is:

1. The system does not know that something like classes even exists.
2. The plugin can override and register any dependency.

src/Squidex.Infrastructure/ComplexQuerySchemaValidationException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace Squidex.Infrastructure
```
SebastianStehle: I try to keep the Infrastructure free from the Squidex domain, Squidex in this case is more the company name ;)

pushrbx: I see. So this class should go somewhere else?

SebastianStehle: Yes, to the namespace with the Interfaces

SebastianStehle:
I would load Modules, not the Interfaces

src/Squidex.Infrastructure/ComplexQuerySchemaValidationException.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public class ComplexQuerySchemaValidationException : Exception
```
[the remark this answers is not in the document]
pushrbx: I have added this exception to the filter, so when the code throws this exception it will produce a 500 status code with an error message.
So the purpose was that I wanted to have something to say "you are returning the wrong schema in your plugin/query, fix it!"

pushrbx:
Changed accordingly.
What's the correct namespace for `ComplexQuerySchemaValidationException`?
`Squidex.Domain.Apps.Read.Contents.CustomQueries`?

src/Squidex.Infrastructure/ComplexQuerySchemaValidationException.cs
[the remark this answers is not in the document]
SebastianStehle: For a 500 you are fine, don't do anything. But when you want a specific error code / message you should add your exception handler.

SebastianStehle:
Some more feedback:

* Can you use IAppEntity whenever you use AppName, because you get access to more infos.
* IQueryModule and IQueryProvider are the same interface. I would do it like the https://github.com/Squidex/squidex/blob/master/src/Squidex.Infrastructure/CQRS/Events/CompoundEventConsumer.cs
* I isolated the different schemas (EDM, OpenApi/Swagger, GraphQL) from the Rest of the Code. QueryArgumentOption violates this rule. The reason is there are so many schemas.
* Try not to use the namespaces in code directly, plz.. e.g. CustomQueries.IQueryProvider 
* `public Task<(ISchemaEntity Schema, long Total, IReadOnlyList<IContentEntity> Items)> Execute(ISchemaEntity schema, QueryContext context, IDictionary<string, object> arguments)` is a little bit strange because it looks like input schema can be different than output schema which should not happen. I think we should just assume that the implementation is correct. There is no way to really test it.

pushrbx:
> `public Task<(ISchemaEntity Schema, long Total, IReadOnlyList<IContentEntity> Items)> Execute(ISchemaEntity schema, QueryContext context, IDictionary<string, object> arguments)` is a little bit strange because it looks like input schema can be different than output schema which should not happen. I think we should just assume that the implementation is correct. There is no way to really test it.

There is a check in the controller:
```cs
var contents = await query.Execute(schema, context, HttpContext.Request.Query.ToDictionary(x => x.Key, x => (object)x.Value));
if (contents.Schema != schema)
{
    throw new ComplexQuerySchemaValidationException("The query has returned an invalid schema.", query.Name);
}
```
-----------
I'll need to think a bit more about the `QueryArgumentOption`.

SebastianStehle:
Yes, but this fix is not useful, because where is the difference between the assumption that the returned content is as expected or that the returned schema is as expected. ;) ... you could also write a filter method that removes all fields that are not valid for the schema. So basically run the validator for the content. This would be more safe but a little bit too much.

pushrbx:
Sorry for being pain in the butt. :smile: 
- What if we just let it throw an exception? When the `Execute` implementation returns an invalid schema it will work in the `ContantApi` but it won't in `graphql`. But I have a hunch that this would be some sort of anti-pattern. 
- I guess for now it is going to be enough to change the signature of `Execute` from `Task<(ISchemaEntity Schema, long Total, IReadOnlyList<IContentEntity> Items)> ` to `Task<IReadOnlyList<IContentEntity>>`. What do you think?

SebastianStehle:
Yes,this was my idea too. Let it fail.

SebastianStehle:
I think you should change the response type a little bit in the ComplexQueryController and not return the total, only a list of items. Because now the total is not valid anymore. You also have to make the change in the SwaggerGenerator.

pushrbx:
I have a strange error now after I've applied the changes on swagger:
```
"An error occurred while deserializing the Clients property of class Squidex.Domain.Apps.Read.MongoDb.Apps.MongoAppEntity: Required element 'Permission' for property 'Permission' of class Squidex.Domain.Apps.Read.MongoDb.Apps.MongoAppEntityClient is missing."
```
Probably it's not related to my changes on swagger, rather I haven't tested it for a while. Also I'm testing with my existing db on Azure, I configured the local development version to connect to the db in azure.
image

pushrbx:
Nvm, my db needs migration.

SebastianStehle:
Just restart Event consumers

pushrbx:
I still have an error after restarting event consumers:
```
Cannot apply $addToSet to a non-array field. Field named 'rd' has a non-array type null in the document _id: "f7ae5bee-8f9f-4197-8c87-2fd56754b997"
```

SebastianStehle:
Have you cleared content?

pushrbx:
I haven't. I'll do that now.

pushrbx:
Swagger shows this now:
squidex-cq-swagger

SebastianStehle:
Awesome

what is the overall status?

pushrbx:
github you are drunk... It has put the wrong timestamp on my comment.
It is done now, it needs a review.

SebastianStehle:
Hi, just wanted to say that I have not forgotten your PR. Will merge it this weekend. I have another PR with a lot of refactorings, which I would like merge first.

SebastianStehle:
See: https://github.com/Squidex/squidex/pull/158

PR #147, opened 2017-10-20
Hotfix - published schemas filter fix for swagger

pushrbx:
Since the last mongodb schema changes under the name `performance improvements` the IsPublished field has been moved inside the `SchemaDocument` field, and that one is the one being updated, not the one outside of `SchemaDocument` field. This pr is a quick fix, probably the proper fix would be to remove the redundant `IsPublished` property from `ISchemaEntity`.


pushrbx:
uhh, i might have to squash the commits sometime. :D

SebastianStehle:
this was a bug it is fixed afaik: https://github.com/Squidex/squidex/blob/master/src/Squidex.Domain.Apps.Read.MongoDb/Schemas/MongoSchemaEntity.cs

pushrbx:
then I close this one.

PR #152, opened 2017-10-22
Refactoring simple model

SebastianStehle:


neeasade:
Could you elaborate on what this PR does a bit? was the goal to simplify models and relations of schema to content?

SebastianStehle:
Yes, it is related to this one. The goal was to make all rich "value objects", e.g. schema more stupid and serialize, so that we can serialize and share them in an easier way, without getting a dependency heal. The goal is to migrate to Orleans, so that Squidex become easier to scale.

Issue #117, opened 2017-09-20
Event Consumer Improvement


SebastianStehle:
The current event consumer system has some issues and we should deal with it.

### Implementation
* Each Collection is an EventConsumer / Subscriber and independent.
* Each EventConsumer has a position, which can be resetted.

### Problems
* Long operations - e.g. imports - for one app also slow down the other app.
* Not scaleable
* Dependencies between event consumers. E.g. App -> Schemas -> Content and App -> Assets
* Slow because of a lot of parallel queries to the event store and deserialization.
* User does not get a notification when there is a problem.

### Requirements
* Apps should not affect each other.
* Better scaleability.
* Maintainance mode, special operations to rebuild the system.
* User should be notified.
* No dependencies between event consumers.
* Faster.

### Idea

1. Each event gets a partitioning key, in our case the AppId to query events by this partitioning key.

2. Each app gets one denormalizer. Denormalizers subscribe to the events for their app as long as they are active. When a denormalizer does not handle an event within 10 minutes, it becomes inactive and will be removed from the list of active denormalizers. This reduces the number of active subscriptions that exist at the same time, because only active denormalizers will have a subscription.

3. The denormalization manager subscribes to all events and ensure that the denormalizer for the current event is active. The denormalizer does not have to handle and deserialize events, therefore it should be much fast than all denormalizers.

3. The denormalization manager (=manager) collects exceptions and shows them in the UI. The administrator can switch to maintenance mode. Then all denormalizers will be stopped. When the user wants to reset the denormalizers the manager creates a single subscription to handle all events together. Otherwise we would have a large range of denormalizers that pull MongoDB at the same time. Once the reset has been completed, the manager sets all denormalizers to active that have received a new event in the meantime.

### Tasks

- [x] Extract the retry logic for event subscriptions. #149 
- [x] Extract logic from domain objects. #148 
- [x] Ensure that all commands are also serializable. #150
- [ ] Introduce partition field and migration tool.
- [ ] Implement new denormalizers.
- [ ] Migrate denormalizers to Orleans or Akka Clusters.

PR #158, opened 2017-11-03
Custom Queries

SebastianStehle:


SebastianStehle:
@pushrbx I checked your PR and I have seen several issues, that I have also not recognized before.

1. The QueryProvider must be an async method, because I would like to read the queries from the DB.
2. The caching of graphql structures can be problematic with the IQueryProvider.
3. I would like to make queries that can also create compound results everything for the home page: News + Blog posts + XXX. Therefore the coupling the schema is not helpful.
4. Because we cannot predict the result, we can also not use it with graphql, especially not with this implementation that is very type safe.

I made a lot of changes to your code by pulling from your branch first and then making the changes step by step. I see a lot of value in you code but it makes it too complicated now, especially with our mid-term plan: https://github.com/Squidex/squidex/issues/117

PR #159, opened 2017-11-06
Allow Media/Asset Inside Rich Text Editor

yuvalsade:
Adds an option to add an asset/external media inside the rich text editor


SebastianStehle:
Can we add a search bar to the dialog?

SebastianStehle:
screencapture-cloud-squidex-io-app-groupl-content-avtogari-textblocks-07d972ab-8a94-4856-b209-861370e365fa-assets-1509960021990

We can also show the asset tool side by side. I think it would be easier if we just support drag and drop because then we profit also from future enhancements. e.g. we are planing asset folders at the moment and would get a problem if we have to implement this twice.

yuvalsade:
basically its possible to add media from the assets folder or from a URL, would you like to have two buttons, one for external media, and one for opening assets side by side for drag and drop?

SebastianStehle:
Yes, two buttons would be great :)

sowobm:
added two buttons on the rich text editor. One for external images and the other which displays the assets manager - where users can just drag and drop the assets in the rich text editor

SebastianStehle:
Awesome, can you revert the changes in these files please:

 * Squidex.csproj
* asset.component.html
* asset.component.ts (hideIcons)

I also do not understand what you need the messages for, I don't need them for the assets editor ;)

The AssetDropHandler is a little bit strange, because it is a service with a dependency to a Pipe. I would place this single method in the asset component.

sowobm:
Alright

- Squidex.csproj, asset.component.html, asset.component.ts have been cleaned from old code leftovers
- I also removed the AssetDropHandler as a standalone service and moved the functionality into the AssetService. I am not sure it belongs there though. It will be hard to have in the asset-component because rich-editor.component depends on it to build the urls while inserting assets into the editor

As for the messages, I am using them to notify the rich-editor.component when to display it's drop zone for asset insertion, when an asset is dragged from the editor. The events are fired in the dragStart and dragEnd on an asset.

sowobm:
I am going to take care of improvement #3 on the next commit, it would be fantastic if you get round to finishing the #1 & #2. 

Sure, it will be great if you could quote us as contributors from http://www.cpmstars.com :)

SebastianStehle:
wow, nice :)

SebastianStehle:
I will take over now, thank you for following my architecture and realizing my feedback :)

sowobm:
It is quite a well written code base ;-) the kind that makes you want to follow.

PR #164, opened 2017-11-09
Fixing error message responses.

### LINKED ISSUE #163: Error messages always showing Failed to make the request
Error messages are no longer showing details from ErrorDto. All errors come back with Failed to make the request.  

Example recreate: When trying to create a duplicate app.

dsbegnoche:
Fixes #163


dsbegnoche:
I am pretty sure that the object will always be in JSON format, but added the object check just in case.

PR #167, opened 2017-11-11
Fixes #166

### LINKED ISSUE #166: Permission null when click Squidex Icon from All Apps
When inside of an app, clicking on the Squidex Icon to return to all apps throws a JS error: Cannot read property 'permission' of null, in left-menu-component.html.  This is caused by ctx.app being null.  Was able to fix this by reverting view to reference permission property in typescript that is set on ngOnInit().

This uncovers another null reference on the Dashboard under the same scenario. This is caused by the subscription to ctx.appChanges in dashboard-page.component.ts firing and the app being null.  The only way that I was able to resolve this was to unsubscribe from the ctx.appChanges. However, this does not feel like the proper solution?

I can create a PR with the fixes I have, but wanted feedback first as unsure if resolution is the proper fix.

dsbegnoche:
.filter on selected app resolves ctx.app.permission null error.
.first() on appchanges() resolves dashboard app.name null error.

### DIFF
--- a/src/Squidex/app/shared/components/app-context.ts
+++ b/src/Squidex/app/shared/components/app-context.ts
@@ -31,7 +31,7 @@ export class AppContext implements OnDestroy {
     }
 
     public get appChanges(): Observable<AppDto> {
-        return this.appsStore.selectedApp;
+        return this.appsStore.selectedApp.filter(a => !!a).map(a => a);
     }
 
     public get appName(): string {
--- a/src/Squidex/app/shared/components/app-context.ts
+++ b/src/Squidex/app/shared/components/app-context.ts
@@ -31,7 +31,7 @@ export class AppContext implements OnDestroy {
     }
 
     public get appChanges(): Observable<AppDto> {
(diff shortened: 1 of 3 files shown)


SebastianStehle:
I don't see why you need the filter. 
I had the same problem when I used the AppComponentBase approach:
 https://github.com/Squidex/squidex/blob/03606fe5c339fbedce16f7af782b18dd2f1cccde/src/Squidex/app/shared/components/app.component-base.ts

dsbegnoche:
It's not required, but prevents app from being set to null.  Without it, I had to have the code from the first commit, where left-menu-component set permission in ngOnInit(). Filter seemed cleaner to me.
https://github.com/Squidex/squidex/pull/167/commits/ae555ff0fafc4196c62c559070cba8f08934b395

SebastianStehle:
I have solved it in another way. I think it is more what i would expect from the AppContext and a null-App is a valid state.

dsbegnoche:
Thank you!

PR #170, opened 2017-11-15
fix bug.

maooson:
Menu title should be changed follow the app selected.


SebastianStehle:
Can you merge in the current master branch? Otherwise the build will fail I think.

But looks good. My AppContext change introduced so much issues :(

SebastianStehle:
Btw: You could also solve it purely in the Template: https://alligator.io/angular/ngif-new-features-angular4/

PR #177, opened 2017-11-16
Allow Dropdown and Radio fields to be empty if they are not required

kwyss88:
Currently, dropdown and radio fields are invalid until a value is selected even if they are not required and doesn't allow the user to save the content item. This update allows dropdown and radio fields to be submitted without selecting a value if the field is not required.


SebastianStehle:
Awesome. Thank you very much.

PR #197, opened 2017-11-28
Number/Radio with Default Value always invalid

dsbegnoche:
The PR from yesterday broke other types. Instead of casting to allowed values to string, should have forced the actual value to be correct type. My apologies for not testing other types first :(


SebastianStehle:
Shit happens :D

PR #198, opened 2017-11-29
Display Default Values in UI

dsbegnoche:
Display Default values in UI on New Items.  

Also, fix for empty Radio/Dropdown numbers via schemas.service.ts and content-field.component.html

### DIFF
--- a/src/Squidex/app/shared/services/schemas.service.ts
+++ b/src/Squidex/app/shared/services/schemas.service.ts
@@ -802,272 +802,272 @@ export class SchemasService {
         const url = this.apiUrl.buildUrl(`api/apps/${appName}/schemas`);
 
         return HTTP.getVersioned<any>(this.http, url)
-            .map(response => {
-                const body = response.payload.body;
+                .map(response => {
+                    const body = response.payload.body;
 
-                const items: any[] = body;
+                    const items: any[] = body;
 
-                return items.map(item => {
-                    const properties = new SchemaPropertiesDto(item.properties.label, item.properties.hints);
+                    return items.map(item => {
+                        const properties = new SchemaPropertiesDto(item.properties.label, item.properties.hints);
 
-                    return new SchemaDto(
-                        item.id,
-                        item.name, properties,
-                        item.isPublished,
-                        item.createdBy,
-                        item.lastModifiedBy,
-                        DateTime.parseISO_UTC(item.created),
-                        DateTime.parseISO_UTC(item.lastModified),
-                        new Version(item.version.toString()));
-                });
-            })
-            .pretifyError('Failed to load schemas. Please reload.');
+                        return new SchemaDto(
+                            item.id,
+                            item.name, properties,
+                            item.isPublished,
+                            item.createdBy,
+                            item.lastModifiedBy,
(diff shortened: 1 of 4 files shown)


src/Squidex/app/shared/services/schemas.service.ts
```diff
@@ -753,272 +802,272 @@ export class SchemasService {
...
-                .do(() => {
-                    this.analytics.trackEvent('Schema', 'FieldDeleted', appName);
-                })
-                .pretifyError('Failed to delete field. Please reload.');
+            .do(() => {
+                this.analytics.trackEvent('Schema', 'FieldDeleted', appName);
+            })
+            .pretifyError('Failed to delete field. Please reload.');
     }
 }
```
dsbegnoche: Damn. Formatting issue for some reason. Will commit revert of that :(

src/Squidex/app/shared/services/schemas.service.ts
```diff
@@ -333,6 +337,8 @@ export abstract class FieldPropertiesDto {
     public abstract formatValue(value: any): string;
 
     public abstract createValidators(isOptional: boolean): ValidatorFn[];
+
+    public abstract getDefaultValue(): any;
```
SebastianStehle: A 5 minute request: Can you make this method not abstract, then we could get rid of a lot of dummy implementations.

src/Squidex/app/shared/services/schemas.service.ts
```diff
@@ -473,6 +491,17 @@ export class DateTimeFieldPropertiesDto extends FieldPropertiesDto {
 
         return validators;
     }
+
+    public getDefaultValue(): any {
+        if (this.calculatedDefaultValue != null) {
+            let now = new Date();
+            if (this.calculatedDefaultValue === 'Today') {
+                now = new Date(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate());
```
SebastianStehle: The implementation is different from the server implementation where we would only return dates for Today (no Time). Would be good to create some tests for it (and use DateTime class).

src/Squidex/app/shared/services/schemas.service.ts
dsbegnoche: Good point. Fixed.

SebastianStehle:
@dsbegnoche Have you seen my other comment with DateTime Field?

dsbegnoche:
Just now. Forgot about the DateTime class, so looking into that.

SebastianStehle:
I improved the tests, also in the API code (C#). Basically we made the same mistake:

I think this is a bad test, because it is hard to say, what you actually expect:

      expect(dto.getDefaultValue()).toEqual(now.toUTCStringFormat('ddd, DD MMM YYYY hh:mm:ss Z'));

I made the same mistake in the C# code:

      Assert.Equal(Now.ToString().Substring(10), DefaultValueFactory.CreateDefaultValue(field, Now));

Actually I meant SubString(0, 10) or "yyyy-MM-dd". I made it more explicit.

dsbegnoche:
Great point. Thank you for the improvements

SebastianStehle:
Thanks for the PR :)

PR #203, opened 2017-12-05
Custom app patterns

dsbegnoche:
Ability to customize Patterns on a per app basis.

1.  Default patterns are created on App Create from appsettings.json
2.  Patterns can be modified in App Settings 
3. Custom patterns can set Name, Pattern and Default Message
4. Patterns must be unique in both Name and Pattern

Questions remaining on implementation:
1. I removed UI Controller in favor of AppPatterns controller. Patterns now behave much like clients/languages etc...  However, this does mean that existing apps will not have patterns showing.  What is the best way to handle these existing apps and getting default patterns created?  A migration process or another option?
2. We want to update any pattern in use on a field when it is updated via Settings.  So far, the only implementation we have of this is pretty ugly.  It requires looping through all the schemas to find any fields with the pattern applied, then update the field with the new value.  It seems their has to be a better way, but currently have not discovered that. Any ideas would be appreciated.   Additionally, is this behavior something that makes sense in Squidex or would it be best for updates to not touch the existing schemas?


src/Squidex.Domain.Apps.Core.Model/Apps/AppPattern.cs
```diff
@@ -0,0 +1,18 @@
+﻿// ==========================================================================
+//  AppPattern.cs
+//  Squidex Headless CMS
+// ==========================================================================
+//  Copyright (c) Squidex Group
+//  All rights reserved.
+// ==========================================================================
+namespace Squidex.Domain.Apps.Core.Apps
+{
+    public class AppPattern
```
SebastianStehle: Classes must be immutable in the Core Model.

src/Squidex.Domain.Apps.Core.Model/Apps/AppPatterns.cs
```diff
@@ -0,0 +1,63 @@
...
+//  Copyright (c) Squidex Group
+//  All rights reserved.
+// ==========================================================================
+using System.Collections.Immutable;
+using System.Diagnostics.Contracts;
+using Squidex.Infrastructure;
+
+namespace Squidex.Domain.Apps.Core.Apps
+{
+    public sealed class AppPatterns : DictionaryWrapper<string, AppPattern>
```
SebastianStehle: Classes must be immutable.

src/Squidex.Domain.Apps.Write/Apps/Commands/UpdatePattern.cs
```diff
@@ -0,0 +1,49 @@
...
+using System;
+using System.Collections.Generic;
+using Squidex.Domain.Apps.Core.Schemas;
+using Squidex.Infrastructure;
+
+namespace Squidex.Domain.Apps.Write.Apps.Commands
+{
+    public sealed class UpdatePattern : AppAggregateCommand, IValidatable
+    {
+        public string OriginalName { get; set; }
```
SebastianStehle: Why do I need the original values?

src/Squidex/Areas/Api/Controllers/Apps/AppPatternsController.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <remarks>
+        /// Gets all configured regex patterns for the app with the specified name.
+        /// </remarks>
+        [HttpGet]
+        [Route("apps/{app}/patterns/")]
+        [ProducesResponseType(typeof(AppPatternDto[]), 200)]
+        [ApiCosts(1)]
+        public IActionResult GetPatterns(string app)
+        {
+            return Ok(App.Patterns.Values?
```
SebastianStehle: A temporary variable would make debugging much easier.

src/Squidex/Areas/Api/Controllers/Apps/Models/AppPatternDto.cs
```diff
@@ -1,16 +1,15 @@
...
+//  AppPatternDto.cs
 //  Squidex Headless CMS
 // ==========================================================================
 //  Copyright (c) Squidex Group
 //  All rights reserved.
 // ==========================================================================
-
 using System.ComponentModel.DataAnnotations;
 
 namespace Squidex.Areas.Api.Controllers.UI.Models
```
SebastianStehle: Wrong namespace

src/Squidex/Areas/Api/Controllers/Apps/AppPatternsController.cs
SebastianStehle: Can we make a fallback here to the config, when the pattern list is null? Would prevent us from having migrations.

src/Squidex/Areas/Api/Controllers/Apps/AppPatternsController.cs
```diff
@@ -0,0 +1,131 @@
...
+using Squidex.Infrastructure.Reflection;
+using Squidex.Pipeline;
+
+namespace Squidex.Areas.Api.Controllers.Apps
+{
+    /// <summary>
+    /// Manages and configures app patterns.
+    /// </summary>
+    [ApiAuthorize]
+    [MustBeAppOwner]
```
SebastianStehle: I think developer should be enough

src/Squidex/app/features/settings/settings-area.component.html
```diff
@@ -32,12 +32,19 @@ <h3 class="panel-title">Settings</h3>
                         <i class="icon-angle-right"></i>
                     </a>
                 </li>
+                <li class="nav-item">
```
SebastianStehle: Please check permissions here

src/Squidex.Domain.Apps.Write/Apps/Guards/GuardAppPattern.cs
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    error(new ValidationError("Pattern can not be empty.", "Name"));
+                }
+
+                if (!patterns.ContainsKey(command.OriginalName.ToLower()))
+                {
+                    error(new ValidationError("Pattern not found.", nameof(command.OriginalName)));
+                }
+
+                if (patterns.Where(x => !x.Key.Equals(command.OriginalName, StringComparison.OrdinalIgnoreCase))
```
SebastianStehle: I dont understand this line. Why do you need a dictionary when you have alreaded looped over all the items?

src/Squidex.Domain.Apps.Write/Apps/Guards/GuardAppPattern.cs
```diff
@@ -0,0 +1,95 @@
...
+                    .ToDictionary(x => x.Key, y => y.Value).ContainsKey(command.Name.ToLower()))
+                {
+                    error(new ValidationError("Pattern name is already assigned.", "Name"));
+                }
+
+                if (patterns.Where(x => !x.Key.Equals(command.OriginalName, StringComparison.OrdinalIgnoreCase))
```
SebastianStehle: I dont understand this line. Why do you need a dictionary when you have alreaded looped over all the items?

dsbegnoche:
1 - Agreed, I don't recall the though process behind using name as keys.  Caused alot of extra code and is unclear.  Will add Id's to use as Keys
2 - Are you saying that instead of updating/modifying/deleting a single pattern, instead we would update the entire collection at once?
3 - Sorry about that.  I recently started using VSCode, and apparently have some formatting problems with it.  I do not seem to have them with VS itself, so need to figure out what is going on there.  I forgot to go back and fix the formatting before pushing :(

SebastianStehle:
1. Depends on 2.
2. Yes, I mean, how many items are there. We talk about very small payloads.
3. No problem, just recognized it.

dsbegnoche:
Added Id for Key, and cleaned up some code.  As for saving the all patterns at once, I am not sure on doing that.  One thing I mentioned in initial PR, was that we are looking to update any field that is using a pattern when the pattern is updated.  First, I am not sure on best way to handle that, but I think that saving the entire list at once will make that more difficult to handle.

SebastianStehle:
Looks very good for me :) ... I am working on the CQRS stuff at the moment and will it merge it after that. For 100% next week.

dsbegnoche:
Thanks Sebastian.  Do you have any thoughts on how we could solve the update requirement on our end?  Right now it just loops through all schemas to find files using the pattern, and updates them.  I don't like that method though.  My only other thought right now though is to maybe have mongo do it with a filter and update?

SebastianStehle:
The fastest approach would be to use  a secondary index

SebastianStehle:
Hi @dsbegnoche 
 
Because there are so many conflicts I have integrated the code step by step by copying files and checking them. Also found some bugs ;)

https://github.com/Squidex/squidex/pull/208

PR #204, opened 2017-12-07
Adding the option to use Google Maps for the geolocation map.

Avd6977:


Avd6977:
It still has the geocoder option on the backend. That way an address can be passed into the API and it'll try to geocode if they're using Google Maps.

PR #214, opened 2018-01-03
Fix

SebastianStehle:


src/Squidex.Infrastructure.GetEventStore/EventSourcing/GetEventStore.cs
```diff
@@ -54,9 +54,30 @@ public IEventSubscription CreateSubscription(IEventSubscriber subscriber, string
             return new GetEventStoreSubscription(connection, subscriber, projectionHost, prefix, position, streamFilter);
         }
 
-        public Task GetEventsAsync(Func<StoredEvent, Task> callback, CancellationToken cancellationToken, string streamFilter = null, string position = null)
+        public async Task GetEventsAsync(Func<StoredEvent, Task> callback, CancellationToken cancellationToken, string streamFilter = null, string position = null)
         {
-            throw new NotSupportedException();
+            var streamName = await connection.CreateProjectionAsync(projectionHost, prefix, streamFilter);
+
+            var sliceStart = ProjectionHelper.ParsePosition(position) ?? -1;
```
pushrbx: this should be
```cs
var sliceStart = ProjectionHelper.ParsePosition(position) ?? 0;
```
Otherwise it throws an exception `System.ArgumentOutOfRangeException: start should be non-negative.` when the position was null.

After changing this I ended up with the following exception:
`System.InvalidOperationException: EventStoreConnection 'ES-52bd9eec-4649-4669-b352-bcc651
cf1686' is not active.`

SebastianStehle: Thats strange

pushrbx: If my ES version is 4.0.1.0, would it cause problems?

pushrbx: Found the problem: see issue #215

SebastianStehle: Can you help me to test this PR?

pushrbx:
I'm testing this now.

pushrbx:
Also I will put back the -1:
```cs
var sliceStart = ProjectionHelper.ParsePosition(position) ?? -1;
```
and test it that way as well.

pushrbx:
One more thing is missing I think: The GetEventStore should test the projection host as well. I had the wrong IP in the config and it gave me a `System.Threading.Tasks.TaskCanceledException`, which doesn't indicate the actual problem at all. You can only assume something from the stack trace:
```
at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at EventStore.ClientAPI.Transport.Http.HttpAsyncClient.<>c__DisplayClass10_0.<RequestSent>b__0(Task`1 task)
```

pushrbx:
With the -1 in it still gives me `System.ArgumentOutOfRangeException: start should be non-negative.`

SebastianStehle:
Ups, sorry

SebastianStehle:
@pushrbx Just saw your comment regarding the test of the ProjectionHost...I think you are right, good point. Will add it today

PR #216, opened 2018-01-05
Fix for migration in case of GetEventStore

pushrbx:
With GetEventStore the migration doesn't work, for two reasons:
- the calculated stream name is wrong
- the stream needs to be resolved

I'm not sure why this happened. I've seen new projections popping up in my ES instance, so I don't know how this was originally, but I managed to make it work with this change.


SebastianStehle:
Your fix helped me, I was just too lazy to merge it ... I also improved the check of the projection manager as you suggested.

Issue #219, opened 2018-01-15
Create ElasticSearch RuleActionHandler


pushrbx:
Create a rule action handler for elasticsearch to let the users create rules to add, update, delete items in the elasticsearch index on content events. (+assets)

Requirements:
- Configuration of elasticsearch host (should be able to accept an array for clusters), password.
- Create indexes for schemas (?)
- Create type names for schemas (?)

Related links:
https://www.algolia.com/
(add this as an extra)

*(?) - not sure about it.

pushrbx:
We have already done this. So this should be closed
#228 https://github.com/Squidex/squidex/commit/7aba54c7defeb585e01d9ddafed28308a622869c

PR #228, opened 2018-01-21
Rule action handler for ElasticSearch

pushrbx:
See issue #219.

WIP PR for now.
### Missing things:
- ~~when `searchIndexing` is not configured (currently optional), the rule creation should give back an error message if the user tries to create rule with and elasticsearch action handler.~~
- icon and colour for the elasticsearch action handler button in the rule wizard. @SebastianStehle it's your task to figure these out 😄 
- ~~Algolia support~~
- Creation of indexes.
- Creation of types.
- Asset indexing

### Poor solutions:
- `RuleActionDtoFactory` has to be more robust, couldn't come up with a better solution
- `RuleAction` now has a Validator property, which shouldn't belong there, but I was too lazy to come up with a proper solution. Any suggestions?
- Not compatible with cloud.squidex.io (global elasticsearch host config)

### Added dependency:
- Low level client to ElasticSearch. (ElasticSearch.Net)

### Worth noting:
- Currently the rule action system is not extendable enough. This PR might have to address this issue.

image

image


pushrbx:
This PR might need squashing too.

SebastianStehle:
Thx a lot. I commented some basic stuff, but the main problem is that the implementation only works for self hosted scenario. For the cloud I need an implementation where the client or the clients are managed by the action itself. The action must also accept an endpoint URL and credentials.

pushrbx:
Will do those when I have time.

SebastianStehle:
Great. I am also open for discussions about the architecture, but I would prefer to create some more actions and then have another discussion how to improve them. We have to get an understanding but the best abstraction is.

pushrbx:
One of the solutions for the problem that currently this is only good for self-hosted envs is this:
image

But what if this action handler would be able to index to any kind of search engine?
In that case this form would have a fourth field, which would define the search engine to use.
So the abstraction for search engines would be inside the action handler.

The other solution I could come up is some sort of settings hub for rules where you could register settings for action handlers and select these settings when you create an action handler which requires one of these settings. Obviously this requires more work compared to the solution above.

SebastianStehle:
I would like to make 2 actions, because they would get individual colors and icons and son on.

I don't get the problem with the screenshot. Can you not add something for authentication also?

SebastianStehle:
I was also thinking about a more generic approach to auto create the UI. But I think I would make a Javascript SDK for it, so that we would have a simple approach to download action handlers from somewhere.

pushrbx:
I forgot about authentication, sorry. 😄  I'll add it soon.

pushrbx:
image

image

The checkboxes are invisible because my current xfce theme isn't compatible with firefox. (linux) \o/

SebastianStehle:
Nice, I think we should also support a placeholder for the schema name for index name and type name.

SebastianStehle:
I will make an action handler for algolia in the mean time. Then we can figure out if they have something in common and refactor it.

pushrbx:
How do I get the schema name in a component? Is it exposed somewhere? (angular)

SebastianStehle:
In the component? I mean we could just add it as hint 

e.g. default value for type name is `$SCHEMA_NAME` and in the action handler we just make a string.Replace

pushrbx:
Alright, I'll add that then.

SebastianStehle:
I thought this might be interesting as well:

https://github.com/Squidex/squidex/issues/236

SebastianStehle:
@pushrbx I refactored the string replacement to a custom class: https://github.com/Squidex/squidex/blob/feature-algolia/src/Squidex.Domain.Apps.Core.Operations/HandleRules/RuleEventFormatter.cs

pushrbx:
I'll merge that one in then.

a977122  - now it works, it sends the data to ES. But it needs further testing, I couldn't create proper mappings for such tests yet. (I always get parser errors from ES)

SebastianStehle:
And another one: https://github.com/Squidex/squidex/pull/238

pushrbx:
I merged in your changes, and it seems on my end the serialization is broken. I will investigate it sometime this week.

SebastianStehle:
I promised it to a client and therefore I have written it bymyself. Already had a lot of "infrastructure" from other actions, so it was jsut a little bit of copy and paste.

pushrbx:
Sorry that I couldn't finish it in time. :( But basically this one was ok, just not really tested. Could you please link here your version?

SebastianStehle:
Sure: https://github.com/Squidex/squidex/commit/7aba54c7defeb585e01d9ddafed28308a622869c

Issue #196, opened 2017-11-28
OData for Assets


dsbegnoche:
Would like to have OData query capabilities for assets so that I can query on any property of an asset.  I.E. Sort by size or type, filter by Mime Types or IsImage. 
This would also help clean up the signature of the GetAssets action. 

Are their any technical reasons why this currently is unavailable?

SebastianStehle:
No, just priorities.

PR #235, opened 2018-01-26
Odata for Assets

dsbegnoche:
https://github.com/Squidex/squidex/issues/196

### DIFF
--- a/src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
+++ b/src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
@@ -15,7 +15,6 @@ using MongoDB.Driver;
 using Squidex.Domain.Apps.Entities.Assets;
 using Squidex.Domain.Apps.Entities.Assets.Edm;
 using Squidex.Domain.Apps.Entities.Assets.Repositories;
-using Squidex.Domain.Apps.Entities.Assets.State;
 using Squidex.Domain.Apps.Entities.MongoDb.Assets.Visitors;
 using Squidex.Infrastructure;
 using Squidex.Infrastructure.MongoDb;
@@ -44,7 +43,6 @@ namespace Squidex.Domain.Apps.Entities.MongoDb.Assets
                     .Ascending(x => x.AppId)
                     .Ascending(x => x.IsDeleted)
                     .Ascending(x => x.FileName)
-                    .Ascending(x => x.MimeType)
                     .Descending(x => x.LastModified));
         }
 
@@ -71,10 +69,11 @@ namespace Squidex.Domain.Apps.Entities.MongoDb.Assets
                 .Find(Filter.In(x => x.Id, ids))
                 .SortByDescending(x => x.LastModified);
 
-            var assetEntities = await find.ToListAsync();
-            var assetCount = await find.CountAsync();
+            var assetEntities = find.ToListAsync();
+            var assetCount = find.CountAsync();
+            await Task.WhenAll(assetEntities, assetCount);
 
-            return ResultList.Create(assetEntities.OfType<IAssetEntity>().ToList(), assetCount);
+            return ResultList.Create(assetEntities.Result.OfType<IAssetEntity>().ToList(), assetCount.Result);
         }
 
         public async Task<IAssetEntity> FindAssetAsync(Guid id)
@@ -90,7 +89,7 @@ namespace Squidex.Domain.Apps.Entities.MongoDb.Assets
         {
             try
             {
-                var model = modelBuilder.BuildEdmModel(new AssetState());
+                var model = modelBuilder.EdmModel;
 
                 return model.ParseQuery(query);
             }
--- a/src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
+++ b/src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
@@ -6,6 +6,7 @@
 // ==========================================================================
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.OData;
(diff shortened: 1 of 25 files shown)


src/Squidex.Domain.Apps.Entities/Contents/ContentOperationContext.cs
```diff
@@ -106,7 +107,20 @@ public Task EnrichAsync()
 
         private async Task<IReadOnlyList<IAssetInfo>> QueryAssetsAsync(Guid appId, IEnumerable<Guid> assetIds)
         {
-            return await assetRepository.QueryAsync(appId, null, new HashSet<Guid>(assetIds), null, int.MaxValue, 0);
```
SebastianStehle: Please use this Method instead:

Task<IResultList<IAssetEntity>> QueryAsync(Guid appId, HashSet<Guid> ids);

A little bit stupid to format to odata first and then parse it again ;)

src/Squidex.Domain.Apps.Entities/Assets/Edm/EdmModelBuilder.cs
```diff
@@ -0,0 +1,61 @@
...
+//  Copyright (c) Squidex Group
+//  All rights reserved.
+// ==========================================================================
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.OData.Edm;
+using Squidex.Infrastructure;
+
+namespace Squidex.Domain.Apps.Entities.Assets.Edm
+{
+    public class EdmModelBuilder : CachingProviderBase
```
SebastianStehle: There is no need to cache it, it will never change. We can just create a private readonly instance for that.

src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
```diff
@@ -34,44 +41,40 @@ protected override Task SetupCollectionAsync(IMongoCollection<MongoAssetEntity>
...
-                    .Descending(x => x.State.LastModified));
+                    .Ascending(x => x.AppId)
+                    .Ascending(x => x.IsDeleted)
+                    .Ascending(x => x.FileName)
+                    .Ascending(x => x.MimeType)
+                    .Descending(x => x.LastModified));
         }
 
-        public async Task<IResultList<IAssetEntity>> QueryAsync(Guid appId, HashSet<string> mimeTypes = null, HashSet<Guid> ids = null, string query = null, int take = 10, int skip = 0)
+        public async Task<IResultList<IAssetEntity>> QueryAsync(Guid appId, string query = null)
```
SebastianStehle: For IContentRepository we pass in the ODataUriParser. We should be consistent here.

src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
```diff
@@ -34,44 +41,40 @@ protected override Task SetupCollectionAsync(IMongoCollection<MongoAssetEntity>
         {
             return collection.Indexes.CreateOneAsync(
                 Index
-                    .Ascending(x => x.State.AppId)
-                    .Ascending(x => x.State.IsDeleted)
-                    .Ascending(x => x.State.FileName)
-                    .Ascending(x => x.State.MimeType)
-                    .Descending(x => x.State.LastModified));
+                    .Ascending(x => x.AppId)
```
SebastianStehle: We have to review this index.  I think we can remove MimeType.

src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
```diff
@@ -34,44 +41,40 @@ protected override Task SetupCollectionAsync(IMongoCollection<MongoAssetEntity>
...
-            var assetItems = Collection.Find(filter).Skip(skip).Limit(take).SortByDescending(x => x.State.LastModified).ToListAsync();
-            var assetCount = Collection.Find(filter).CountAsync();
+        public async Task<IResultList<IAssetEntity>> QueryAsync(Guid appId, HashSet<Guid> ids)
+        {
+            var find = Collection
+                .Find(Filter.In(x => x.Id, ids))
+                .SortByDescending(x => x.LastModified);
 
-            await Task.WhenAll(assetItems, assetCount);
+            var assetEntities = await find.ToListAsync();
```
SebastianStehle: Can you use Task.WhenAll? => 2x Performance.

src/Squidex.Domain.Apps.Entities.MongoDb/Assets/Visitors/ConstantVisitor.cs
```diff
@@ -0,0 +1,67 @@
+﻿// ==========================================================================
```
SebastianStehle: You should get errors from roslyn about the header ;)

SebastianStehle:
Why not building it around IQueryable interface? Then you can use it more easily for other storage providers. For IContentEntity it was not possible, because of the highly dynamic nature of this data structure, but for IAssetEntity it should be straight forward.

src/Squidex.Domain.Apps.Entities.MongoDb/Assets/MongoAssetRepository.cs
dsbegnoche: IContentRepository has the ContentQueryService over it, which is where converting string to ODataUriParser exits.  IAssetRepository doesn't have that abstraction.  I will add an AssetQueryService to do this.

SebastianStehle: You are right, then just keep it like that.

SebastianStehle:
@dsbegnoche Just foiund this: https://docs.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/supporting-odata-query-options

SebastianStehle:
Should be easy to extract that to an extension method / methods.

dsbegnoche:
I get the idea around IQueryable, but not sure if it would be all that straightforward?  Also, using WebAPI OData Query I believe would make the Content OData queries be different than Assets? For instance, to return paging with the WebAPI2 you must specify $count. It also supports does not support $search..

SebastianStehle:
Afaik you can enable or disable certain  features. I think it is worth a try.

PR #241, opened 2018-02-02
Fix SecurityDefinitions in Swagger helper

mjhpour:
According to OAS2 the key name in "security" should match with
the key name in "SecurityDefinitions" to apply intended security
to the specified scope.


SebastianStehle:
Thank you

mjhpour:
@SebastianStehle Nop.

PR #244, opened 2018-02-05
Added strict port definition in program

blakecodes:
On startup the program selects a dynamic port to run from by default, I've added a strict port definition to the program.cs file to solve this.


SebastianStehle:
I think it is VS that does it.

PR #257, opened 2018-03-06
Pipeline Unit Tests

Avd6977:


SebastianStehle:
Awesome. From what I see now I am confused why you use Moq AND FakeItEasy...

tests/Squidex.Tests/GlobalSuppressions.cs
```diff
@@ -0,0 +1,14 @@
+﻿// ==========================================================================
```
SebastianStehle: Can you remove this file?

Avd6977:
Some entities I got the error that it couldn't be "Faked" because it wasn't an interface.

Done

SebastianStehle:
But why do you want to mock ActionDescriptor? It is just a dto. And for HttpContext you can use DefaultHttpContext

Avd6977:
For HTTP Context I wanted to be able to control certain pieces of the request/response and I couldn't get that working without mocking. I did go ahead and change the Action Descriptor to the DTO though. Maybe I'm just missing something but I can't get it working without mocking HTTP Context and the response/request.

PR #296, opened 2018-05-23
MongoDB GridFS support

razims:


src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
```diff
@@ -0,0 +1,219 @@
...
+        public async Task CopyAsync(string name, string id, long version, string suffix,
+            CancellationToken ct = default(CancellationToken))
+        {
+            try
+            {
+                var file = GetFile(name);
+
+                using (var stream = new MemoryStream())
+                {
+                    await bucket.DownloadToStreamByNameAsync(file.Name, stream, cancellationToken: ct);
```
SebastianStehle: Can we not just rename the asset?

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
```diff
@@ -0,0 +1,219 @@
...
+            catch (IOException ex)
+            {
+                if (!directory.Exists)
+                {
+                    throw new ConfigurationException($"Cannot access directory '{directory.FullName}'", ex);
+                }
+            }
+        }
+
+        public string GenerateSourceUrl(string id, long version, string suffix)
```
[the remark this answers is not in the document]
razims: agree, but i would keep files locally anyway

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
razims: no support from IGridFSBucket, unfortunately :(

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
```diff
@@ -0,0 +1,219 @@
...
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new AssetNotFoundException($"Asset {id}, {version} not found.", ex);
+            }
+        }
+
+        public Task UploadAsync(string name, Stream stream, CancellationToken ct = default(CancellationToken))
```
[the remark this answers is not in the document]
SebastianStehle: Why? What are the cleanup rules? When do we delete them? What happens when you have multiple nodes? What is the advantage over simple folders?

GenerateSourceUrl is not that important. We can just return "UNSUPPORTED". The main reason is for debugging.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
```diff
@@ -0,0 +1,219 @@
...
+            => DeleteCoreAsync(GetFile(name));
+
+        public Task DeleteAsync(string id, long version, string suffix)
+            => DeleteCoreAsync(GetFile(id, version, suffix));
+
+        private async Task DeleteCoreAsync(FileInfo file)
+        {
+            try
+            {
+                var filter =
```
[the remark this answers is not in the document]
razims: should i push a change here ?

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
SebastianStehle: Or at least a version where you do not keep everything in memory.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
SebastianStehle: If you want to keep a local copy I think it would make sense to make something like an "Aggregate" Asset store. Like a Multi Level Cache.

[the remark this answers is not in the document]
razims: expression bodies -> as you want Sir :)
GenerateSourceUrl  -> haven't digged too much into logic, if you say so then fine

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
razims: Rename != Copy

razims: no :) , copy means you have two copies at the end

razims: it would make sense to use Rename if you'll rename IAssetStore.CopyAsync to IAssetStore.RenameAsync

SebastianStehle: Right

razims:
One more comment. Since you use Autofac, i see no reason to use Guard.NotNull for injected objects. Autofac has built in check for this. It's better to use JetBrains.Annotation like [NotNull] and if you use Resharper you'll get notified during coding.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
SebastianStehle: The real reason of the GenerateSourceUrl was to provide the url where you can download the file, e.g. when you make a public bucket you could the direct source url to bypass squidex. The implementation for the FileStorage does not really make sense.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
SebastianStehle: Would be great.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
```diff
@@ -0,0 +1,219 @@
...
+            catch (GridFSException ex)
+            {
+                throw new AssetNotFoundException($"Asset {name} not found.", ex);
+            }
+        }
+
+        public async Task DownloadAsync(string id, long version, string suffix, Stream stream,
+            CancellationToken ct = default(CancellationToken))
+        {
+            var file = GetFile(id, version, suffix);
```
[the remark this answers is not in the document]
SebastianStehle: As I said: if you want to improve performance it would make more sense to implement something like a CachingAssetStore that has an inner IAssetStore. Because you can use the caching for all other storages as well then. And if you want to cache it you have also to clean it up. Otherwise you could have a lot of files and you have no idea where all your disk space is gone.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
SebastianStehle: Yes, as I said you are right. I was thinking too much about the use case here, where it does not really matter if it is copy or rename because the source will be deleted anyway ;)

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: my idea was to keep local files and have GridFS as reserve store. so if i deploy application in a new machine it should be able to sync files from reserve place as GridFS. I would keep the possibility to serve static files here, bypassing squidex as you mentioned.

SebastianStehle:
I do not use Autofac. I like to keep the a class independent of the infrastructure, so that I get proper exceptions. I do not use Resharper anymore because it was too annoying to keep the code styles and warnings consistent across the environments. Much better with Roslyn now :)

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: CachingAssetStore  -> agree

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: sounds good

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: where you want to cache the files? Memory?? You'll cache them on disk anyway, locally

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: May be IFileProvider abstraction would be better than IEnumerable. What you think? This would allow to store nested way too as bonus, if required of course. 

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/file-providers?view=aspnetcore-2.1#file-provider-abstractions

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: We can have a flag, keepLocal=true/false

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
SebastianStehle: Depends how much cache levels you want.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
SebastianStehle: The IFileProvider is readonly, unfortunately. I mean something like this: https://github.com/aspnet/FileSystem/blob/dev/src/FS.Composite/CompositeFileProvider.cs

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
```diff
@@ -141,16 +141,13 @@ public Task DeleteAsync(string id, long version, string suffix)
...
             try
             {
-                var filter =
-                    Builders<GridFSFileInfo>.Filter.And(
-                        Builders<GridFSFileInfo>.Filter.Eq(x => x.Filename, file.Name)
-                    );
-                using (var cursor = await bucket.FindAsync(filter))
+                file.Delete();
+                using (var cursor = await bucket.FindAsync(Builders<GridFSFileInfo>.Filter.And(
+                    Builders<GridFSFileInfo>.Filter.Eq(x => x.Filename, file.Name)
```
SebastianStehle: I still do not see why you have to loop. Can you not use use: `bucket.DeleteAsync(file.Name)`?

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
SebastianStehle: Thats fine for me too.

razims:
failed - no space left on device :(

SebastianStehle:
Yes, no problem. The next commit will work again ;)

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
razims: Unfortunately no support from IGridFSBucket to delete fine by name currently.

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: what about the case when we have set keepLocal=false earlier and then call GenerateSourceUrl to get "static" url?  Create locally, then serve?

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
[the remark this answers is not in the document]
razims: CompositeFileProvider.cs just a set of IFileProviders. For reading fine

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
SebastianStehle: Is the name not the same like Id?

SebastianStehle: Here it looks like it could be anything: https://mongodb.github.io/mongo-csharp-driver/2.4/apidocs/html/T_MongoDB_Driver_GridFS_GridFSBucket_1.htm

razims: nope, it is not

razims: it is object identifier "Id" from GridFSFileInfo 
http://api.mongodb.com/csharp/current/html/T_MongoDB_Driver_GridFS_GridFSFileInfo.htm

SebastianStehle: It says that there is a generic alternative, so you could use IGridFSBucket<string> and use FileName (Name+version+suffix) as Id.

SebastianStehle: yes

razims: it will also overcomplicate bucket.UploadFromStreamAsync(file.Name, stream, cancellationToken: ct);

SebastianStehle: Perhaps I have another version, but for me it looks like

`bucket.UploadFromStreamAsync(file.Name, file.Name, stream, ...);`

https://mongodb.github.io/mongo-csharp-driver/2.4/apidocs/html/M_MongoDB_Driver_GridFS_GridFSBucket_1_UploadFromStreamAsync.htm

src/Squidex.Infrastructure.MongoGridFs/Assets/MongoGridFsAssetStore.cs
```diff
@@ -73,13 +73,13 @@ public string GenerateSourceUrl(string id, long version, string suffix)
             {
                 var file = GetFile(name);
 
+                file.CopyTo(GetPath(id, version, suffix));
+
                 using (var stream = new MemoryStream())
                 {
-                    await bucket.DownloadToStreamByNameAsync(file.Name, stream, cancellationToken: ct);
-                    await bucket.UploadFromStreamAsync(file.Name, stream, cancellationToken: ct);
+                    await bucket.DownloadToStreamAsync(file.Name, stream, cancellationToken: ct);
```
SebastianStehle: Can be improved with

 ```
using (var readStream = await bucket.GetDownloadStreamAsync(file))
{
	using (var writeStream = await bucket.GetUploadStreamAsync(file.Name))
	{	
		var buffer = new byte[4096]; // Should be equivalent to options.BlockSize
		var bytesRead = 0;
		
		while ((bytesRead = await readStream.ReadAsync(buffer, 0, buffer.Length)) 
		{
			await writeStream.WriteAsync(buffer, 0, bytesRead);
		}
	}
}
```

razims: yes, right

razims: but where you found GetDownloadStreamAsync?

SebastianStehle: sry, OpenDownloadStreamAsync

razims: where you took options.BlockSize from ?

SebastianStehle: I mean ChunkSizeBytes

Jaben:
This isn't going to be merged into the project @SebastianStehle ?

SebastianStehle:
I thought the development has kind of stopped because I had the impression that some of the discussions were still open / not resolved.

razims:
actually not, we're already using it internally works great. I thought you were not interested 😀 to merge it. Anyway please clarify, what is required and what you don't like. I'll fix/push in next days.

SebastianStehle:
I just had a look to the code again and I am not sure If I remember everything about the local file handling. In a multi-node environment the files do not always exist, but the error handling does not reflect it.

razims:
@SebastianStehle if you mean storing the copies of images locally, initially i thought it may increase the throughput and response time. But then i discovered that it adds additional "cache" level and increases complexity. It should generally be the task of consumer, whether he/she wants to cache the result or not. I'll remove the local file handling.

SebastianStehle:
That would be great. If requested I would implement a cache for ALL storage providers.

Furthermore: 

1. Please move the class to the Infrastructure.MongoDb project
2. Please to not swallow exceptions in Initialize()

razims:
@SebastianStehle agreed

SebastianStehle:
Btw: I hope you now the new forum: https://support.squidex.io/

razims:
@SebastianStehle yep, you invited once

Jaben:
I'm using it and quite happy with it so far... nice work @razims . One thing I noticed doesn't work is the backup -- assets are all empty.

SebastianStehle:
Could also be the backup system. I will test it when I merge it.

razims:
you didn't wait for me 😀

SebastianStehle:
I misunderstood you and thought it is already done. When I merged it locally I realized that I can just finish within half an hour and just completed it.

razims:
okay, fine 👍🏻

PR #305, opened 2018-07-08
Alpine linux docker image

pushrbx:
I've changed the dockerfile to build an alpine image, and also I've changed the .csproj file of the Squidex project to be bound to the 2.1.1 dotnet runtime. This way the image size will be around 163MB, which could be reduced to around 90MB but I didn't want to risk it.
@SebastianStehle what are the tests you used to conduct to verify it is stable with such changes?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,21 +28,25 @@ RUN dotnet restore \
  && dotnet test tests/Squidex.Tests/Squidex.Tests.csproj
 
 # Publish
-RUN dotnet publish src/Squidex/Squidex.csproj --output /out/ --configuration Release
+RUN dotnet publish src/Squidex/Squidex.csproj --output /out/alpine --configuration Release -r alpine.3.7-x64
 
 #
 # Stage 2, Build runtime
 #
-FROM microsoft/dotnet:2.1.0-aspnetcore-runtime
+FROM microsoft/dotnet:2.1-runtime-deps-alpine
 
 # Default AspNetCore directory
 WORKDIR /app
 
+# add libuv
+RUN apk add --no-cache libuv \
+&& ln -s /usr/lib/libuv.so.1 /usr/lib/libuv.so
+
 # Copy from build stage
-COPY --from=builder /out/ .
+COPY --from=builder /out/alpine .
 
 EXPOSE 80
 EXPOSE 33333
 EXPOSE 40000
 
-ENTRYPOINT ["dotnet", "Squidex.dll"]
+ENTRYPOINT ["./Squidex"]
--- a/src/Squidex/Squidex.csproj
+++ b/src/Squidex/Squidex.csproj
@@ -6,6 +6,7 @@
     <PackageId>Squidex</PackageId>
     <PreserveCompilationContext>true</PreserveCompilationContext>
     <TargetFramework>netcoreapp2.1</TargetFramework>
+    <RuntimeFrameworkVersion>2.1.1</RuntimeFrameworkVersion>
     <TypeScriptToolsVersion>Latest</TypeScriptToolsVersion>
     <TypeScriptCompileBlocked>true</TypeScriptCompileBlocked>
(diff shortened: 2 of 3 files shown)


SebastianStehle:
Great :) I only see manual testing for now.

SebastianStehle:
I think it does not make that much sense to use a self contained deployment with the .net runtime image. Because the runtime is not used anyway.

SebastianStehle:
Why is libuv needed? I thought it has been removed with 2.1.

SebastianStehle:
Ignore my last question. I have not realized that the socket implementation is not stable yet.

pushrbx:
> I think it does not make that much sense to use a self contained deployment with the .net runtime image. Because the runtime is not used anyway.

All the tutorials I've checked online mentioned that this has to be setup like this. I will try look in to alternatives, because I don't have too much knowledge on this.
But basically are you saying that there has to be a way where we omit the `-r` option from the dotnet publish arguments and use the dotnet cli in the alpine image to start Squidex?

SebastianStehle:
My understand is that -r adds all the runtime assemblies (like libcorecrl.so) to the output folder. Basically they exist two times now:

1. As part of the deployment
2. As part of the base image

So it makes no sense to combine it.

I also found some more things to reduce the size:

1. NSwag is too big. Way to big: https://github.com/RSuter/NSwag/issues/1446
2. Get rid of CodeAnalysis assemblies

image

pushrbx:
I gave it a shot today.
I've changed the docker file to the following:
```
#
# Stage 1, Prebuild
#
FROM squidex/dotnet:2.1-sdk-chromium-phantomjs-node as builder

COPY src/Squidex/package.json /tmp/package.json

# Install Node packages 
RUN cd /tmp && npm install

COPY . .

WORKDIR /

# Build Frontend
RUN cp -a /tmp/node_modules /src/Squidex/ \
 && cd /src/Squidex \
 && npm run test:coverage \
 && npm run build:copy \
 && npm run build
 
# Test Backend
RUN dotnet restore \
 && dotnet test tests/Squidex.Infrastructure.Tests/Squidex.Infrastructure.Tests.csproj \ 
 && dotnet test tests/Squidex.Domain.Apps.Core.Tests/Squidex.Domain.Apps.Core.Tests.csproj \ 
 && dotnet test tests/Squidex.Domain.Apps.Entities.Tests/Squidex.Domain.Apps.Entities.Tests.csproj \
 && dotnet test tests/Squidex.Domain.Users.Tests/Squidex.Domain.Users.Tests.csproj \
 && dotnet test tests/Squidex.Tests/Squidex.Tests.csproj

# Publish
RUN dotnet publish src/Squidex/Squidex.csproj --output /out/ --configuration Release

#
# Stage 2, Build runtime
#
FROM microsoft/dotnet:2.1-aspnetcore-runtime-alpine

# Default AspNetCore directory
WORKDIR /app

# Copy from build stage
COPY --from=builder /out/ .

EXPOSE 80
EXPOSE 33333
EXPOSE 40000

ENTRYPOINT ["dotnet", "Squidex.dll"]
```

The size of the image went up from 168mb to 240mb with this change.

pushrbx:
I'll do more digging if it is bigger than the current one. Atleast this PR needs to produce a smaller image, because otherwise it doesn't make any sense to merge it.

pushrbx:
@SebastianStehle Sorry about the delay on this. Now I had time to look into this. The docker hub is compressing the images, so the size you can see there is the compressed size of the image. So far I was talking about the size what I could see by typing `docker images` in the terminal.
So in uncompressed form, and with the self-contained deployment the alpine image is smaller than the original one:
image

SebastianStehle:
Is the PR ready to be merged?

SebastianStehle:
Looks also good in dockerhub: https://hub.docker.com/r/squidex/squidex/tags/

PR #312, opened 2018-08-22
Add Identity Server as an external login

nvella88:
With reference to [https://support.squidex.io/t/login-using-openid/136/6](the ticket) please find a basic setup of a custom identity server acting as an external resource. Please remember to restore a DB for Identity Server `dotnet ef database update -c ApplicationDbContext`. 

P.S. There is a slight issue regarding confirmation (after Signup) but this also happens with Google Authentication. If I have time I might try to address this problem in a seperate PR.


SebastianStehle:
Can you provide your sample as a separate repository?

nvella88:
The setup is a seperate fork, publicly available. Would this suffice?

SebastianStehle:
I can just use the separate repo.

nvella88:
Was forking a bad idea?

SebastianStehle:
It depends on you. I would never accept the pull request because the identity server is part of the repo. I would keep the identity server separated. There is also keycloak if you need a identity server with UI: https://www.keycloak.org/

And of course Squidex identity.

nvella88:
I understand this is only for testing purposes. In my projects everything will be seprate. I closed the PR because it doe snot make sens but I still hope you can give it a spin to see the behaviour yourself.

Thanks for keycloak and for Squidex Identity, I am learnign a lot from it.

SebastianStehle:
Yes, I will have a look to it. Perhaps we can make a it an official feature to configure an external open id provider through the config.

PR #313, opened 2018-08-29
Rule Pager Broken

dsbegnoche:
Rules pager does not page. Add prev/next actions.


SebastianStehle:
Thank you very much, have you no solved a similar problem before? ;)

PR #317, opened 2018-09-14
Save Query box fixes

dsbegnoche:
Save Query box does not clear out last saved item when opening
Cancel button saves query

### DIFF
--- a/src/Squidex/app/shared/components/search-form.component.html
+++ b/src/Squidex/app/shared/components/search-form.component.html
@@ -91,7 +91,7 @@
             </ng-container>
         
             <ng-container footer>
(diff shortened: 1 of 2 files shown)


SebastianStehle:
Thank you. The "offical" way is to call form.submitCompleted({}) in this case.

SebastianStehle:
Can you fix it?

Btw: Can you write me a testimonial: https://support.squidex.io/t/request-for-testimonials/165 ?

dsbegnoche:
Yes. Will get that updated in a few.

dsbegnoche:
Updated with correct form reset.

PR #319, opened 2018-09-26
Schema Category Behavior Update

Avd6977:
Updating the schema category behavior to:
- Set the open state properly using this.name instead of name
- Expand all categories when searching and return to previous open state when clearing the query
- Hide categories that don't have query results and update the count to match the count found by query


SebastianStehle:
Can you prepare a short gif to show the differences: https://www.screentogif.com/

SebastianStehle:
Thank you very much. Makes sense :)

PR #326, opened 2018-10-09
Pending should be false when changing status

Avd6977:
Fixing a weird state on content
1. Create and publish a content item
2. Make a pending change
3. Select "Unpublish" and the pending data is still there.

If you're unpublished, you shouldn't be able to have a pending version.

### DIFF
--- a/src/Squidex.Domain.Apps.Entities/Contents/State/ContentState.cs
+++ b/src/Squidex.Domain.Apps.Entities/Contents/State/ContentState.cs
@@ -86,6 +86,7 @@ namespace Squidex.Domain.Apps.Entities.Contents.State
         protected void On(ContentStatusChanged @event)
         {
             ScheduleJob = null;
+            IsPending = false;
 
             Status = @event.Status;
 


Avd6977:
You're right. We would need to sync the data and dataDraft in that scenario. Which you don't know if they would be approving or rejecting the pending changes. What are your thoughts on it?

SebastianStehle:
just an else?

```
protected void On(ContentStatusChanged @event)
        {
            ScheduleJob = null;

            Status = @event.Status;

            if (@event.Status == Status.Published)
            {
                Data = DataDraft;
            }
            else
            { 
                DataDraft = Data;
            }
            
            IsPending = false;
        }
```

Avd6977:
Either that or hide the options from them and not allow them to change status without accepting or rejecting the changes. The else makes an assumption that they want to discard the changes.

SebastianStehle:
Or we always set `Data = DataDraft` so you do not loose changes. Actually it does not really matter.

PR #328, opened 2018-10-16
FieldEditWizard

Avd6977:
Allow users to edit the field after they create it in the field wizard. They can also go back and keep adding fields and editing them until they are done with that schema.
fieldeditwizard


Avd6977:
I'm with you on the three green buttons, but also unsure how to solve that.
1) fixed (got bit by auto-format in VSCode)
2) fixed
3) nothing, changed it to isEditing so that could be used instead of step
4) Removed the class that was causing the weird styling

Avd6977:
Is this waiting on anything?

SebastianStehle:
No, I just hoped that I would find a solution for the three buttons problem.

PR #330, opened 2018-10-23
Deactivate rule if disabled

kwyss88:
The change deactivates the rule if a user has disabled it.


SebastianStehle:
Awesome, but the PR fails with some "style errors". You should see them in the IDE

kwyss88:
Sorry about that. I have committed the fix.

PR #331, opened 2018-10-23
Added unit test for disabled rules

kwyss88:


SebastianStehle:
Awesome, but the PR fails with some "style errors". You should see them in the IDE

PR #337, opened 2018-12-22
Fixed isLocked issue

seamys:
This problem occurs when the administrator tries to lock the user and the data is changed, but the UI is never updated.
gif


SebastianStehle:
Thank you. I refactored the user system recently and I broken it :(

seamys:
It's a great project, I learned a lot from you.

SebastianStehle:
Thank you :)

PR #346, opened 2019-02-13
Make orleans ports configurable

mhilgersom:
We're hosting Squidex on Microsoft Azure Webapps. For deployment we're using a blue-green deployment model with different slots within an Azure webapp (http://www.tiernok.com/posts/blue-green-deployment-to-azure-app-services/).

This works like a charm. Only with Orleans this might lead to a situation where both slots are on the same backend server trying to setup a Silo with the same ports. 

Attached change will use an EnvironmentVariable to override the default siloPort and gatewayPort. 

This way we can give the Orleans cluster different port numbers per slot.

### DIFF
--- a/src/Squidex/Config/Orleans/OrleansServices.cs
+++ b/src/Squidex/Config/Orleans/OrleansServices.cs
@@ -58,8 +58,8 @@ namespace Squidex.Config.Orleans
             {
                 ["MongoDB"] = () =>
                 {
-                    var siloPort = int.Parse(Environment.GetEnvironmentVariable("Orleans.SiloPort") ?? "11111");
-                    var gatewayPort = int.Parse(Environment.GetEnvironmentVariable("Orleans.GatewayPort") ?? "40000");
+                    int siloPort = int.TryParse(config.GetRequiredValue("orleans:siloPort"), out siloPort) ? siloPort : 11111;
+                    int gatewayPort = int.TryParse(config.GetRequiredValue("orleans:gatewayPort"), out gatewayPort) ? gatewayPort : 40000;
 
                     hostBuilder.ConfigureEndpoints(Dns.GetHostName(), siloPort, gatewayPort, listenOnAnyHostAddress: true);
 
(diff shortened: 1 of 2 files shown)


SebastianStehle:
Thank you for your contribution.

Please use  the configuration system for that.

Either add a custom `MyOrleansOptions`  class and use the options system or use the config object. The ASP.NET Core configuration system supports multiple sources out of the box including environment variables.

You should also handle the case if a non-integer value was entered with int.TryParse() and document the options by adding them to the appSettings.json.

mhilgersom:
Never knew that environment settings could even override appsettings. Changed to using the config object. And added TryParse().

SebastianStehle:
Mhm, int.TryParse does not make that much sense in combination with GetRequiredValue ...

I also think that the gateway port is not needed, because we do not connect with external gateways to orleans.

SebastianStehle:
Btw: I just realized that in your scenario it would be easier to turn off the clustering: https://github.com/Squidex/squidex/blob/master/src/Squidex/appsettings.json#L180

Sorry for the delay. Changing the port in a blue green scenario does not really help either. Sorry for being late.

mhilgersom:
Turning off clustering (I think you mean setting mode to Development) results in a silo starting at 127.0.0.1:11111. 

The problem we're facing is that both slots seems to be running on the same backend vm. So in the original config they both start a silo on the hostname / localhost at the same port number. Although apps within Azure should be seperated they do interfere with one another resulting in one of both silo's getting in a restart loop. 

I've tested my change with specific port numbers per slot in our environment. This resolve the problem. Both webapps start their silo on a seperate number and they don't seem to interfere with one another. Swapping slots doesn't break any of the silo's. 

Reverting back to the default config or disabling the cluster introduces the errors again. 

We're hosting our apps in a Azure Internal App Service Environment. Perhaps this strange behaviour is caused by this type of environment.

SebastianStehle:
Okay, I will merge it tomorrow. But I would really like to understand the problem. What exactly happens? Without clustering there should be communication over these ports at all.

PR #348, opened 2019-02-22
Fixed null eventsfilter issue

seamys:
Fixed null eventsfilter issue


SebastianStehle:
Thank you very much :)

PR #350, opened 2019-03-19
Feature/content sorting

oguzhankahyaoglu:
Back-end implementation for content sorting; 

1. an OrderNo field is added to ContentEntity (on for MongoDb)
2. On any insert, new OrderNo value is automatically propogated.

**Things I could not resolve:**
1. UI Implementation: a sortable view should be used and whenever rows sorted, database should be updated by switching related orderno values correspondingly.
2. Migration: for projects that already have some content data, a migration should be applied by updating all contents' OrderNo fields increasingly. I could not be able to understand the migration technology behind the project and I did not want to change the infrastructure for such a need.


SebastianStehle:
I don't get it: What is the purpose of the orderNo field?

SebastianStehle:
Ah, sorry, it is for sorting the items.

PR #353, opened 2019-04-23
fix event panel width

YehudaKremer:
small fix for event panel.
see image: https://drive.google.com/file/d/1Fr65_iV_Zk3LpeW3RWhWra_VahsQaxAb/view?usp=sharing


SebastianStehle:
It is actually by design.

YehudaKremer:
In the image you see the title "Roles" is cut by the panel (mark with red circle).
Any why, is not big deal.. :)

PR #357, opened 2019-05-04
Maximum callstacksize exceeded

awinnen:
Hi i discovered an issue with control-errors component and rich text editor.

It will throw error "maximum call stack size exceeded". Changes to the tinyMCE editor are lost upon saveing sometimes.

To reproduce this issue:

1. Have at least two languages active in your app
2. Create Schema with localizable field with rich text editor
3. Create new content for that schema.
4. Modify rich text for language 1
5. Switch language to language 2
6. Modify rich text for language 2
7. Switch language to language 1 again
8. Modify rich text for language 1
9. Click save Button
reproduce error

I found out, that this issue seems to be caused by the control-errors component decorating the markAsTouched function of the formControls. But when the formcontrol changes on language switch, the original function is not restored.


SebastianStehle:
Thank you very much.

PR #358, opened 2019-05-17
Added kafka Extension which can push events to Kafka Topics.

sauravvijay:
This is just a basic pull request which can push squidex events to Kafka broker. 
Now I'm working on extending this functionality where we can configure SchemaRegistry and Avro Schema for the data we want to push to Kafka Topics.


SebastianStehle:
Great, I will wait for your changes. In general I am a little bit skeptical if the rule system makes sense for kafka. There is so much to configure for the connection and you also need a certificate in some cases.

Here is some more feedback:

* The key should be either configurable or the AggregateId, otherwise your events for the some content or asset could be out of order.
* You should cache the connection (producer), otherwise you will create and open a new connection for every event and you would probably run out of sockets, see: https://github.com/Squidex/squidex/blob/master/extensions/Squidex.Extensions/Actions/Algolia/AlgoliaActionHandler.cs
* The producer is relatively heavy weight, therefore I cannot use it in the cloud. It would make sense to create a new project: `Squidex.Extensions.SelfHosted` (or so), with all the stuff that is not cloud ready. I can then disable this plugin for the cloud.
* Perhaps it makes more sense to implement this similar to the rabbitmq event consumer: https://github.com/Squidex/squidex/blob/master/src/Squidex.Infrastructure.RabbitMq/CQRS/Events/RabbitMqEventConsumer.cs.

sauravvijay:
Hi @SebastianStehle , I've implemented the changes requested

SebastianStehle:
I am going to make these changes and test it with another kafka server as well.

SebastianStehle:
Pushed from my PR after I tested it.

PR #360, opened 2019-05-18
Add FTP asset store supports to squidex.

seamys:
Hi @SebastianStehle ,

Add FTP asset store supports to squidex.


seamys:
Hi, 
I'm a little confused about this fail.
image

SebastianStehle:
This should not be related to your issue.

SebastianStehle:
Thanks a lot,

1. You can also add it to Squidex.Infrastructure directly if you want. The benefit is that you can use the ready to use unit tests there: https://github.com/Squidex/squidex/tree/master/tests/Squidex.Infrastructure.Tests/Assets

2. Why do you create a new client each time? I would create the client as a readonly field and then connect once in the InitializeAsync method and also implement IDisposable. It is probably faster.

seamys:
Hi, @SebastianStehle 

Thank you for providing these helpful suggestions. 

For question 2, 

it was implemented as you described.  but, I'm worried if client disconnect for some reason (network issue). the whole function will stop working, and it can not self-healing. 

so, I made a change on it,  Create a ftpClient without connecting. move connecting to every method.
in every method end disconnect. but I'm worried about multi-thread situation.

That is why I made it create a new client each time.  

let me change it.

SebastianStehle:
About question 2: I see, what is the recommendation from SmartFTP guys? Btw: I would also move the least important methods (creating the client) to the bottom of the file.

seamys:
@SebastianStehle 
I'm using FluentFTP (https://github.com/robinrodricks/FluentFTP).  I didn't ask them for recommendations.

but I saw the example.  they all created client each time.
https://github.com/robinrodricks/FluentFTP/tree/master/FluentFTP.Examples

The key code is here:
https://github.com/robinrodricks/FluentFTP/blob/master/FluentFTP/Stream/FtpSocketStream.cs#L814

SebastianStehle:
Seems to make sense then.

Btw: Why do you write your own tests? Just derive from the asset store test.

seamys:
Oh, I got you. asset store test is a basic test.

SebastianStehle:
Yes, exactly. The advantage is that the tests is like a list of requirements. If your store passes all tests it is good to go.

seamys:
Does this PR also need to do in the same way?

SebastianStehle:
Yes please. You have written a unit test, but in my experience it does not prove anything. You have to mock the underlying system a lot and really understand the implementation for the test to work. If you upgrade the package and the behavior changes a little bit, your tests are still green, but it does not work in production. Therefore test base is an integration test and does not run on the build server.

seamys:
@SebastianStehle 

Yes, I thought the test is the unit test (**mock the dependence and focus the unit code**).

SebastianStehle:
Why have you closed it?

seamys:
Sorry, I thought you want to move this to the plugin system.  need to reopen?

SebastianStehle:
I merged your stuff manually.

seamys:
thanks, more know this system, more love it.

PR #361, opened 2019-05-18
Feature/SendGrid

seamys:
Added sendgrid support

### DIFF
--- a/extensions/Squidex.Extensions/Actions/SendGrid/SendGridAction.cs
+++ b/extensions/Squidex.Extensions/Actions/SendGrid/SendGridAction.cs
@@ -21,7 +21,7 @@ namespace Squidex.Extensions.Actions.SendGrid
     {
         [Required]
         [Display(Name = "APIKey", Description = "Sendgrid's API Key.")]
-        [DataType(DataType.Text)]
+        [DataType(DataType.Password)]
         public string APIKey { get; set; }
 
         [Required]
--- a/extensions/Squidex.Extensions/Actions/SendGrid/SendGridAction.cs
+++ b/extensions/Squidex.Extensions/Actions/SendGrid/SendGridAction.cs
@@ -1,0 +1,51 @@
+﻿// ==========================================================================
+//  Squidex Headless CMS
+// ==========================================================================
+//  Copyright (c) Squidex UG (haftungsbeschraenkt)
+//  All rights reserved. Licensed under the MIT license.
+// ==========================================================================
+
+using System.ComponentModel.DataAnnotations;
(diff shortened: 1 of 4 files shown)


SebastianStehle:
Cool stuff, I will definitely integrate it. The question is how. My idea of the plugin system was to allow external contributors like you to provide plugins. The benefit is that

a) You are free to do whatever you want
b) I do not have to maintain the contributions by myself.

What is your point of view about this?

SebastianStehle:
e.g. as an example: I am the co-maintainer for several Orleans extensions:

* https://github.com/OrleansContrib/OrleansDashboard
* https://github.com/OrleansContrib/Orleans.Providers.MongoDB
* https://github.com/OrleansContrib/WebHostCompatibilityLayer

SebastianStehle:
I tested it how to create a plugin from your SendGrid rule and added you to the repository.

seamys:
@SebastianStehle 
Sorry, I still can't understand the idea you describe. is it like plugin market?  like WordPress, able to install plugin online?

I'm using docker image form squidex hub.

SebastianStehle:
Yes, that iss basically the idea. I hope that this enables more developers to contribute to Squidex.

SebastianStehle:
What is your opinion?

seamys:
Oh, After I saw the project that you created the project sendgrid. that is defiantly a good way.
I will use it.

seamys:
Hi, Other questions. 

1. How about the custom editor.  can we do it in the same way?  Custom editor some time need pulling data from squidex by using API.

2. Do we really need to clone the whole project for a single plugin?

SebastianStehle:
1. For custom editors we already have a deployment strategy, you just need a link. But the interface should and must be extended to provide information like the API token and the current user.

2. No, I am also thinking about deploying the Squidex assemblies to NuGet, but I think I have to reorganize the assemblies before to distinguish between interfaces (Squidex.XXX.Abstractions) and implementations. By cloning the whole project you have the advantage that it easy to test your plugin.

seamys:
@SebastianStehle 
Okay, got it

SebastianStehle:
Thank you for your feedback. Please also feel free to say if you dislike my ideas or decisions. Feedback is always great, especially when it comes from valuable contributors like you :)

seamys:
Yes, I completely understand your idea. but do you know when the final version will come out? 
I don't comfortable with the current version. because it makes docker image large and large where we trying to add many plugins.

SebastianStehle:
No, I don't know, but the final version cannot do things smaller. It is the 6-8 MB that is the limit. It is because of how .NET manages dependencies.

seamys:
I see.

PR #368, opened 2019-06-18
Seperate stages for frontend and backend.

SebastianStehle:


SebastianStehle:
@sauravvijay I have cosed this PR for now. Your idea with 2 layers is great, but Squidex is not prepared for that. I think it would better to separate the frontend from the backend from the root folder and then copy them separately. Then the layering would work much better. Perhaps something for next sprint.

PR #369, opened 2019-06-19
Workflow generalization

SebastianStehle:


src/Squidex.Domain.Apps.Core.Model/Contents/StatusChange.cs
```diff
@@ -1,17 +1,16 @@
...
-//  Copyright (c) Squidex UG (haftungsbeschraenkt)
+//  Copyright (c) Squidex UG (haftungsbeschränkt)
 //  All rights reserved. Licensed under the MIT license.
 // ==========================================================================
 
 namespace Squidex.Domain.Apps.Core.Contents
 {
     public enum StatusChange
     {
-        Archived,
```
sauravvijay: why have we removed Archived status?

SebastianStehle: It is not the status. It is the transition from Draft => Archived or from Published to Archived.

But when we generalize the workflow we cannot assume anymore that there is a new Archived state, so I have removed this. Published is a special case, it must always exist.

tools/Migrate_01/OldTriggers/ContentChangedTriggerSchema.cs
```diff
@@ -50,19 +50,9 @@ public ContentChangedTriggerSchemaV2 Migrate()
...
-
             if (SendArchived)
             {
-                conditions.Add($"event.type == '{EnrichedContentEventType.Archived}'");
-            }
-
-            if (SendRestore)
-            {
-                conditions.Add($"event.type == '{EnrichedContentEventType.Restored}'");
+                conditions.Add($"event.status == 'Archived'");
```
sauravvijay: Is this correct?

SebastianStehle: No, should be 'Draft', thx

SebastianStehle: No,  everything is fine. The Change here looks just a little bit strange.

tests/Squidex.Domain.Apps.Entities.Tests/Contents/ContentGrainTests.cs
```diff
@@ -337,7 +365,7 @@ public async Task ChangeStatus_should_create_events_and_update_state_when_archiv
 
             LastEvents
                 .ShouldHaveSameEvents(
-                    CreateContentEvent(new ContentStatusChanged { Status = Status.Archived, Change = StatusChange.Archived })
+                    CreateContentEvent(new ContentStatusChanged { Status = Status.Archived })
```
Ben-Fletcher-UK: Has Archived been removed as a possible state now? I have noticed it's gone from the ContentChangedTriggerHandlerTests.cs but it's still in here.

SebastianStehle: It is part of the default implementation.

PR #384, opened 2019-07-02
Feature/ui options

SebastianStehle:
Introduce settings that are added to Index.html
Option to hide news dialog.
Option to hide onboarding tooltips.
Option to redirect to login page automatically without extra click.

### DIFF
--- a/src/Squidex/Areas/Api/Controllers/UI/Models/UISettingsDto.cs
+++ b/src/Squidex/Areas/Api/Controllers/UI/Models/UISettingsDto.cs
@@ -3,26 +3,12 @@
 // ==========================================================================
 //  Copyright (c) Squidex UG (haftungsbeschränkt)
 //  All rights reserved. Licensed under the MIT license.
-// ==========================================================================
-
-using System.ComponentModel.DataAnnotations;
+// ==========================================================================using System.ComponentModel.DataAnnotations;
 
 namespace Squidex.Areas.Api.Controllers.UI.Models
 {
     public sealed class UISettingsDto
     {
-        /// <summary>
-        /// The type of the map control.
-        /// </summary>
-        [Required]
-        public string MapType { get; set; }
-
-        /// <summary>
-        /// The key for the map control.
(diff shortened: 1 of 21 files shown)


src/Squidex/Areas/Api/Controllers/UI/Models/UISettingsDto.cs
```diff
@@ -3,26 +3,12 @@
 // ==========================================================================
 //  Copyright (c) Squidex UG (haftungsbeschränkt)
 //  All rights reserved. Licensed under the MIT license.
-// ==========================================================================
-
-using System.ComponentModel.DataAnnotations;
+// ==========================================================================using System.ComponentModel.DataAnnotations;
```
Jason-Segal: Should this line be removed since it is commented out?

SebastianStehle: Fixed

src/Squidex/app/shared/guards/must-be-authenticated.guard.spec.ts
```diff
@@ -9,24 +9,25 @@ import { Router } from '@angular/router';
...
     let authService: IMock<AuthService>;
-    let authGuard: MustBeAuthenticatedGuard;
+    let uiOptions = new UIOptions({ map: { type: 'OSM' } });
+    let uiOptionsRedirect = new UIOptions({ map: { type: 'OSM' }, redirectToLogin: true });
 
     beforeEach(() => {
         router = Mock.ofType<Router>();
 
         authService = Mock.ofType<AuthService>();
-        authGuard = new MustBeAuthenticatedGuard(authService.object, router.object);
```
Jason-Segal: Just out of interest, how comes this was moved out of the 'BeforeEach' section if it is used in all tests within this file?

SebastianStehle: It is used with different arguments.

PR #385, opened 2019-07-03
Added EnglishNorway language variant

lars-berger:
A minor addition to the languages available.

### DIFF
--- a/src/Squidex.Infrastructure/Languages.cs
+++ b/src/Squidex.Infrastructure/Languages.cs
@@ -319,6 +319,7 @@ namespace Squidex.Infrastructure
         public static readonly Language EnglishMalaysia = AddLanguage("en-MY", "English (Malaysia)");
         public static readonly Language EnglishNamibia = AddLanguage("en-NA", "English (Namibia)");
         public static readonly Language EnglishNorfolkIsland = AddLanguage("en-NF", "English (Norfolk Island)");
+        public static readonly Language EnglishNorway = AddLanguage("en-NO", "English (Norway)");
         public static readonly Language EnglishNigeria = AddLanguage("en-NG", "English (Nigeria)");
         public static readonly Language EnglishNetherlands = AddLanguage("en-NL", "English (Netherlands)");
         public static readonly Language EnglishNauru = AddLanguage("en-NR", "English (Nauru)");


SebastianStehle:
Hi, the languages are actually auto generated: https://github.com/Squidex/squidex/blob/master/tools/GenerateLanguages/Program.cs 

We should also add it to the generator, otherwise it would get overriden.

lars-berger:
Aha, and the cultures that are generated are either in `language-codes.csv` or come from `CultureInfo.GetCultures(CultureTypes.SpecificCultures)`?

Would it be an OK solution to have a `custom-cultures.csv` that is read from like `language-codes.csv`?

SebastianStehle:
Yes, thats fine for me.

Btw: are you sure, that en-NO exists?

lars-berger:
The use-case is for a homepage with language & country selection. As an example, we operate in Norway and Sweden, and have our content translated in English, Norwegian, and Swedish with some small adjustments for each country (eg. link hrefs, contact information & products available).

The plan is to add the following to the language settings: 
```
en       English
en-SE    English (Sweden) - optional and falls back to English
en-NO    English (Norway) - optional and falls back to English
sv       Swedish
sv-SE    Swedish (Sweden) - optional and falls back to Swedish
...
```
We would write the main content in `en` and `sv`, and for fields that are country-specific we would fill out `en-SE`, `en-NO` etc. In the frontend, `en` and `sv` wouldn't be used, only `en-SE`, `sv-SE`, etc.

I have a 2nd approach in mind. For example, for handling which products are available in a country, an array field could be used and I could have checkboxes to select countries where that product should be available. However, I'm unsure how to handle country-specific fields that are used as link hrefs in our frontend (eg. `/en-se/products`). As more and more country-specific fields are added, I feel like there'll just be an accumulation of hacky solutions.

Perhaps there's a better approach to handling this, but I just haven't found it yet.

SebastianStehle:
We could a method to add languages and then a configuration section for additional languages. I don't want to add everbody's special language to the system ;)

lars-berger:
That's a good point, I can see why you wouldn't want that happening... A way to add custom languages (through the frontend I suppose?) sounds like a great idea!

SebastianStehle:
Yes, this would even better. What I was thinking about is a simple option in the appsettings for now.

lars-berger:
In `src/Squidex/appsettings.json`? To be honest, I'd be happy with absolutely anything. C#/.NET is way out of my realm of knowledge (hadn't touched it before yesterday), but I'd be ecstatic to help out if you ever decide to make this an addition to the frontend.

SebastianStehle:
Yes, exactly. I have created a feature request and will close this PR: https://support.squidex.io/t/option-to-add-additional-language-codes-from-settings/883

Thank you for your valuable feedback.

SebastianStehle:
Implemented: https://github.com/Squidex/squidex/commit/6403aeac8c7677abfccd264a11ea4f0ca0b9fe07

lars-berger:
Awesome, thank you!! Can't wait to try it out.

PR #387, opened 2019-07-18
Reference fields.

SebastianStehle:


src/Squidex/app/shared/services/schemas.service.ts
```diff
@@ -99,19 +100,24 @@ export class SchemaDetailsDto extends SchemaDto {
...
-            if (listFields.length === 0) {
-                listFields = NONE_FIELDS;
-            }
+    private getField(predicate: (field: RootFieldDto) => boolean) {
+        let fields = this.fields.filter(x => predicate(x) && x.properties.isContentField);
 
-            this.listFields = listFields;
-            this.listFieldsEditable = listFields.filter(x => x.isInlineEditable);
+        if (fields.length === 0 && this.fields.length > 0) {
+            fields = [this.fields[0]];
```
Ben-Fletcher-UK: Is there any chance this can be null instead of 0? If so what happens?

SebastianStehle: No chance.

PR #407, opened 2019-08-31
Feature/app infos

SebastianStehle:
Introduces labels, descriptions and icons for apps:

image

image


SebastianStehle:
@Ben-Fletcher-UK You can ignore all changes in other services for rules, schemas, assets and so on. I unified some code and removed duplicate stuff.

PR #410, opened 2019-09-05
ability to build custom extensions without forking

alexisbssn:
My goal here is to be able to have a separate solution for my custom code from the squidex solution. 

It will be needed to publish a new nuget package with every new version of Squidex: `Squidex.Extensions.Dependencies`

It will be needed to edit the docker-compose files to include a volume on the plugins folder
e.g. current volume is `/etc/squidex/assets:/app/Assets` and new volume would be `/etc/squidex/plugins:/plugins`

I'm not sure it's a good thing to place the `plugins` folder in the root directory, but I'm also not sure it's good to put it beside the `.ts` files in `/app/`


SebastianStehle:
In general I think it is a good idea, but I am not keen to merge the implementation to Nuget as well. In my opinion there should be a strict separation between implementation and interfaces and only interfaces and helper functions of course should go into nuget.

alexisbssn:
I agree with you. I tried abstracting away but there is just so much that it ends up not working very well.

One offender was Language, which is difficult to abstract without a lot of code change because of implicit operators that can't go on interfaces.

Another one is RuleActionHandler, exposing generic classes via interfaces is possible but very hard to read and maintain: `IExample<TSelf, TSomething> where TSomething : ISomething where TSelf : IExample<TSelf, TSomething>`

What would you suggest?

SebastianStehle:
Language is just a domain object.

It is a class to class decision. I think one indicator is, whether the class has an external dependency or not. I think what needs to be splitted is 

* Squidex.Domain.Apps.Entities
* Squidex.Domain.Apps.Infrastructure

I could do that.

Btw: I think a volume for docker is not a good idea, it makes scaling impossible and does not work well with kubernetes I think.

alexisbssn:
On the topic of volumes, it does work with k8s just as well as with docker-compose. If there is a concern for performance, docker allows the mounting of a readonly volume, and if that's still not enough we could copy the dll files upon startup to a local folder.

I'll let you split the classes as you have a better understanding of what each one does. I'll use my fork in the meantime. Do you think you will prioritize this soon?

SebastianStehle:
I have created a test, how hard it is to move the implementation detail. I have started with Infrastructure and I am almost done.

Should be an easy task.

PR #420, opened 2019-09-30
Asset RemoveAsync fix

dsbegnoche:
Change RemoveAsync to remove single item by id instead of all assets for app.


SebastianStehle:
THanks. Can you change it to a simple "=="?

dsbegnoche:
Updated to use ==

PR #437, opened 2019-11-05
Enable multi-select of roles for prevent updates

mittulmadaan:


mittulmadaan:
Will merge it later in master.

PR #440, opened 2019-11-08
update terser-webpack-plugin to v 2.2.1

stevenalanstark:
see: https://support.squidex.io/t/can-not-build-with-terser-webpack-plugin-v2-1-2/1307


SebastianStehle:
Thx

PR #447, opened 2019-11-18
Include specific tenant in Microsoft Authentication

mhilgersom:
Microsoft Authentication uses a default endpoint for AzureAD authentication. This make authentication with any AzureAd account possible. 

When you wish to limit this to only users from your own AzureAD you need to specify a specific AzureAD tenant. 

This pull request implements adding an optional tenantId field. 

How to test:
- Create an Azure App registration in Azure Active directory.
- Configure Squidex to use this clientId and clientSecret
- Authenticate and verify access --> when no tenant is specified should work as default.

- Change the Azure App registration so only users from this directory can sign in. 
- Restart Squidex
- Try to sign in --> you should get an error stating that the common endpoint cannot be used to authenticate users for a specific AzureAD. 

- Add your AzureAD tenantId to the settings of squidex
- Restart Squidex
- Authenticate --> this should now work as expected


mhilgersom:
See that the same issue is reported here: https://support.squidex.io/t/microsoft-authentification-for-single-tenant/1065

PR #457, opened 2019-12-02
Feature/proxy forwarded host option

mhilgersom:
This PR uses the X-Forwarded-Host header to set the correct hostname in .Net Core when running behind a reverse proxy. 

Only when the X-Forwarded-Host field is present and the hostname matches the baseUrl property and the enableXForwardedHost option is enabled this header will be used.

With this PR setting the PublicOriging for IdentityServer might not be needed anymore.


SebastianStehle:
Thank you very much. We should make it in a way so that is fully backwards compatible.

mhilgersom:
I agree. I think we should remove the fixed PublicOrigin for IdentityServer and replace it with this method of using the X-Forwarded-Host header. 

This way it only affects users that set explicitly set this new option. And it will only be effective when the supplied X-Forward-Host header matches the hostname of the confugred baseUrl setting.

SebastianStehle:
Will you take care about it?

mhilgersom:
Yes. Busy with over project. Will fix asap.

mhilgersom:
I've updated the PR and tested this one within our environment.

PR #464, opened 2019-12-24
Added configuration to override s3 assets store service url from config

rubvardanyan:
Right now squidex supports the only cloud-based Amazon S3 assets store, but there are already implementations of S3 APIs, such as [minio](https://min.io/).
Here I have added configuration to allow overriding of amazon s3 client service URL, so we can use third party implementations of S3 APIs as an assets store.


SebastianStehle:
Thx

PR #465, opened 2019-12-25
Feature/force path style amazon client config

rubvardanyan:
Right now AmazonS3 config used with its default property values. Some S3 API implementation requires specific configuration. I added only `ForcePathStyle` config, but in future would be good to allow set all configurations for AmazonS3 Client.


SebastianStehle:
Can we create an options class that is easier to extend?

rubvardanyan:
> 
> 
> Can we create an options class that is easier to extend?

Sure, I will move all AmazonS3 configuration stuff to the separate option class.

rubvardanyan:
@SebastianStehle done.

PR #466, opened 2019-12-25
Fixed amazon s3 options binding

rubvardanyan:


rubvardanyan:
Fixed options binding created in previous pull request

PR #502, opened 2020-03-25
Adding selected Content to context form

Anders-Arnfast-CreunaDK:


SebastianStehle:
Because it does not get updated automatically when something is changed, would it not be better to name it "initialContent" or so.

SebastianStehle:
There is also another one: https://github.com/Squidex/squidex/blob/master/frontend/app/features/content/shared/references/content-creator.component.ts

Anders-Arnfast-CreunaDK:
Yes, i agree with the naming i'll change that, InitialContent makes sense, in regards to thttps://github.com/Squidex/squidex/blob/master/frontend/app/features/content/shared/references/content-creator.component.ts isn't it a pre-content component, i mean it looks like there isn't any content to add to the form context, or am i wrong?

SebastianStehle:
Yes, you are right :D...but can you also change formContext to contentFormContext

Anders-Arnfast-CreunaDK:
@SebastianStehle wouldn't it be a potential breaking change who people who might use it?

SebastianStehle:
In general, because OnPush, it might be better to generate a new instance of the formContext...

e.g. with

this.contentFormContext = { ...this.contentFormContext, initialContent: content }

Anders-Arnfast-CreunaDK:
@SebastianStehle  is that what you ment?

SebastianStehle:
Sorry, I just meant this single field in the content-page.component  ... :( 

My fault

PR #510, opened 2020-04-13
Analtyics

MartijnDijkgraaf:
i'm only not quite comfortible with the way i made the uitoptions optional. But i didnt find a good way to mock it because the uioptions class in the configuration file.

I didn't want to extract it from the configuration file and make it an own class. because that would make the pull request too big.

Maybe another time

### DIFF
--- a/frontend/app/app.module.ts
+++ b/frontend/app/app.module.ts
@@ -16,7 +16,6 @@ import { RouterModule } from '@angular/router';
 import { AppComponent } from './app.component';
 
 import {
-    AnalyticsIdConfig,
     ApiUrlConfig,
     CurrencyConfig,
     DecimalSeparatorConfig,
@@ -58,10 +57,6 @@ export function configTitles() {
     return new TitlesConfig(undefined, 'Squidex Headless CMS');
 }
 
-export function configAnalyticsId() {
-    return new AnalyticsIdConfig('UA-99989790-2');
-}
-
 export function configDecimalSeparator() {
     return new DecimalSeparatorConfig('.');
 }
@@ -88,7 +83,6 @@ export function configCurrency() {
         AppComponent
     ],
     providers: [
-        { provide: AnalyticsIdConfig, useFactory: configAnalyticsId },
         { provide: ApiUrlConfig, useFactory: configApiUrl },
         { provide: CurrencyConfig, useFactory: configCurrency },
         { provide: DecimalSeparatorConfig, useFactory: configDecimalSeparator },
--- a/frontend/app/framework/services/analytics.service.ts
+++ b/frontend/app/framework/services/analytics.service.ts
@@ -23,15 +23,14 @@ export const AnalyticsServiceFactory = (uiOptions: UIOptions, router: Router, re
 @Injectable()
 export class AnalyticsService {
     private readonly gtag: any;
-    private readonly analyticsId: AnalyticsIdConfig;
+    private analyticsId: AnalyticsIdConfig;
 
-    constructor(private readonly uiOptions: UIOptions,
+    constructor(private readonly uiOptions?: UIOptions,
         private readonly router?: Router,
         private readonly resourceLoader?: ResourceLoaderService
     ) {
         window['dataLayer'] = window['dataLayer'] || [];
-        this.analyticsId = new AnalyticsIdConfig(this.uiOptions.get('google.analyticsId'));
-
+        this.setAnaltyicsId();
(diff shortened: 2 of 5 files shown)


frontend/app/framework/services/analytics.service.ts
```diff
@@ -12,37 +12,29 @@ import { NavigationEnd, Router } from '@angular/router';
...
+    private analyticsId: AnalyticsIdConfig;
 
-    constructor(analyticsId?: AnalyticsIdConfig, router?: Router, resourceLoader?: ResourceLoaderService) {
+    constructor(private readonly uiOptions?: UIOptions,
+        private readonly router?: Router,
+        private readonly resourceLoader?: ResourceLoaderService
+    ) {
         window['dataLayer'] = window['dataLayer'] || [];
-
+        this.setAnaltyicsId();
```
SebastianStehle, line 33: There is a spelling error.

frontend/app/framework/services/analytics.service.ts
```diff
@@ -53,4 +45,31 @@ export class AnalyticsService {
             value: value
         });
     }
+
+    private configureGtag() {
+        if (this.analyticsId && this.router && this.resourceLoader && window.location.hostname !== 'localhost' ) {
+
```
SebastianStehle, line 51: Can you just remove this line?

frontend/app/app.module.ts
```diff
@@ -16,7 +16,6 @@ import { RouterModule } from '@angular/router';
 import { AppComponent } from './app.component';
 
 import {
-    AnalyticsIdConfig,
```
SebastianStehle, line 19: Can you delete this class?

frontend/app/framework/services/analytics.service.ts
MartijnDijkgraaf: which line? I don't see any highlighted lines. But if you mean "`if (analyticsId && router && resourceLoader && window.location.hostname !== 'localhost') {`" and beneath. I didn't remove it, I just cut it up in separate functions.

frontend/app/app.module.ts
MartijnDijkgraaf: which line? I don't see any highlighted lines. But if you mean "`if (analyticsId && router && resourceLoader && window.location.hostname !== 'localhost') {`" and beneath. I didn't remove it, I just cut it up in separate functions.

frontend/app/framework/services/analytics.service.ts
MartijnDijkgraaf: which line? I don't see any highlighted lines. But if you mean "`if (analyticsId && router && resourceLoader && window.location.hostname !== 'localhost') {`" and beneath. I didn't remove it, I just cut it up in separate functions.

frontend/app/framework/services/analytics.service.ts
MartijnDijkgraaf: fixed it

frontend/app/app.module.ts
MartijnDijkgraaf: That class was only used in the configAnalyticsId function, which returned a AnalyticsConfig class. This functions was used in the providers. I removed that provider because we don't need to provide that to the analytics service. I now provide de UIOptions in the analytics services and initialise the AnalyticsConfig class from there.

I hope this makes sense.

MartijnDijkgraaf:
Im quite new to the GitHub reviewing and pull request system, so If somethings not alright please inform me about it.

frontend/app/framework/services/analytics.service.ts
SebastianStehle: The empty line after the if

frontend/app/app.module.ts
SebastianStehle: Yes, therefore we can delete AnalyticsIdConfig

MartijnDijkgraaf: yes, so its okay now? because the import is deleted? 

Or do you mean the whole class that is defined in configurations.ts? Because that class is still getting used.

SebastianStehle: But where do we use it? I think we should get rid of this class.

MartijnDijkgraaf: I use it in the analytics.service, but your right, we can get rid of it. It’s a little bit overkill to create a class just to hold a string.

I’ll fix it tomorrow.

PR #513, opened 2020-04-14
Updated MongoDBDriver to 2.10.3 which fixes DNS resolution bug

mvacha:
There is a bug in DnsClient library used by MongoDBDriver used by Squidex that causes DNS resolution failure when running in AKS because of Microsoft's changes in coredns responses (https://github.com/Azure/AKS/issues/1526).

Root cause have been fixed in DnsClient library and a new version of MongoDBDriver also have been released - fixing this bug:
https://jira.mongodb.org/browse/CSHARP-2968?focusedCommentId=2966977&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2966977


SebastianStehle:
Thank you very much

PR #514, opened 2020-04-17
Enabled HTTPS in dev mode with self signed certificate

sauravvijay:


backend/src/Squidex/Program.cs
```diff
@@ -61,6 +63,16 @@ public static void Main(string[] args)
                 })
                 .ConfigureWebHostDefaults(builder =>
                 {
+                    builder.ConfigureKestrel((context, serverOptions) =>
+                    {
+                        if (context.Configuration.GetValue<bool>("devMode:enable"))
```
SebastianStehle, line 68: Why do we need this? Shouldn't be there the development mode already?

backend/src/Squidex/appsettings.json
```diff
@@ -1,4 +1,9 @@
 ﻿{
+  "devMode": {
+    "enable": true,
```
SebastianStehle, line 3: We cannot set this to true, it would be a breaking change.

backend/src/Squidex/Program.cs
```diff
@@ -61,6 +63,16 @@ public static void Main(string[] args)
...
+                        {
+                            serverOptions.Listen(
+                                IPAddress.Any,
+                                context.Configuration.GetValue<int>("devMode:port"),
+                                listenOptions => listenOptions.UseHttps("squidex-dev.pfx", "password"));
```
SebastianStehle, line 73: Is this a path or a cert name? If it is a path, it would not work

backend/src/Squidex/Program.cs
sauravvijay: this is required so that self signed certificate is only served in development mode.

sauravvijay: we can probably use ASPNETCORE_ENVIRONMENT variable to check if it's running in dev mode and then server certificate.

backend/src/Squidex/appsettings.json
sauravvijay: this property will be removed.

backend/src/Squidex/Program.cs
sauravvijay: this is not a path. Its Personal Information Exchange file which is created along with cert.

backend/src/Squidex/Program.cs
SebastianStehle: Yes, but I mean there is already a development in ASP.NET Core, e.g.

https://github.com/Squidex/squidex/blob/835274396f58f9b9f8f11bab8956639337a97a5b/backend/src/Squidex/Areas/Frontend/Startup.cs#L41

We can introduce an additional flag, but at least we should check for both.

if (context.HostingEnvironment.isDevelopment() || ...)

backend/src/Squidex/Program.cs
SebastianStehle: But the docs say

```
//
        // Summary:
        //     Configure Kestrel to use HTTPS.
        //
        // Parameters:
        //   listenOptions:
        //     The Microsoft.AspNetCore.Server.Kestrel.Core.ListenOptions to configure.
        //
        //   fileName:
        //     The name of a certificate file, relative to the directory that contains the application
        //     content files.
        //
        //   password:
        //     The password required to access the X.509 certificate data.
        //
        // Returns:
        //     The Microsoft.AspNetCore.Server.Kestrel.Core.ListenOptions.
        public static ListenOptions UseHttps(this ListenOptions listenOptions, string fileName, string password);
```

backend/src/Squidex/appsettings.json
SebastianStehle: If we don't need it than we can also hard code the port in my opinion.

backend/src/Squidex/Program.cs
sauravvijay: yes pfx file can be used as a cert file as well. It's kept under the same directory hence used directly with name. Have tested it and works like a charm

PR #515, opened 2020-04-23
WIP: Translation

MartijnDijkgraaf:
Hey Sebastian,

I've created a start the the possibility to translate Squidex. I only translated the home-page because I wanted to see what you thoughts about it are.

For this features I made use of [ngx-translate](https://github.com/ngx-translate/core). In the backed I added some options, the first options is the language which the CMS will use, currently you can only choose between English and Dutch. The second option is the options to let te browser make the choice, also here it can only choose between English and Dutch.

The last option 'languageFolder' give the user the possibility to enter a url to a folder where the json files are stored. for example, is the user wants a German translations, he has to change the language to 'de' and add and url to the folder where there is a de.json and en.json file available. That en.json is for the fallback (maybe we can fix that later in something more user friendly).

But that's wat I wanted to show you, im not sure if you should merge it already into the master, but I couldn't find another way to let you review and see the code.

Kind regards,
Martijn


SebastianStehle:
Thank you for very much, I have some questions / comments:

1. Why do you use ngx-translate and not angular itself? I think angular has the helpful feature to generate the translation files.

2. I think we should keep the frontend stuff out of the backend. It was a painful process to seperate them ;)

MartijnDijkgraaf:
Hey, No problem, In the end it benefits us all.

1. I looked into i18n from Angular and the ngx-translate. The main problem everbody had with i18n was that you needed multiple builds. Thats not a problem if you want to support a fixed number of languages (en, fr,de,nl). But at the moment you need to add another one, it's getting really messy. 

But maybe I can make another pull request with the i18n from Angular, then we can see the real difference.

2. You mean the options in appsettings.json? or the locale folder? The appsetting.json settings can we maybe move to the GUI, maybe add a flag option.
if we add a flag options, the i18n option can work, because then we have a fixed number of languages. 

I will try and make another pull request this weekend (no strings attached). With the i18n one and maybe the separation of the frontend en backend stuff.

Also, im trying to learn Squidex and Angular on the go, so if you have any comments or tips, please let them know. Even if it sucks :D.

SebastianStehle:
About 1: I was no aware about the problem with angular i18n. You are right, that sucks.

Is there any translation tool for ngx-translate? If not we could write one. A little bit of regex would be enough.

About 2: Mainly the ngx-folder

MartijnDijkgraaf:
1. There isn't a tool like i18n has, that it recognises the translatable text.
But if we fill the English json file manually we can edit the other language files with  [Babeledit](https://www.codeandweb.com/babeledit. (it's 35 euro incl VAT). 

2. Maybe its the lack of coffee but I don't see the ngx-folder, or do you mean wwwroot > **locale**.

SebastianStehle:
1. I know, how hard can it be?

I would write a custom i18n attribute, then loop over all the templates and use a regex (`i18n="(?<Key>.*)">(?<Text>.*)<\/span>`) To get the texts and put them in the translation file.

This also has the other advantage, that when a translation is not complete, you can but in the english version as well.

We also need an automated solution for the initial work, because going over all files is boring and time intensive.

2. Yes, I mean the locale folder.

MartijnDijkgraaf:
I just read something about the i18n from Angular, that on the long term it will make ngx-translate useless. So I think for the benefit of Squidex and its future, i should make a poc with Angular's i18n.

Then we can see the benefits of both and make a choice.

2. I tried to put that folder initially in the frontend, but I couldn't make the folder available for the 'server'? I knowledge about Webpack is not that big, but ill try it again.

3. Is there a way to make the `nom make` faster? the waiting times are a pain in the ass at this moment.

Summary:
I'll make a PoC with i18n so we can see the benefits of both choices. I keep you updated.

MartijnDijkgraaf:
I just read something about the i18n from Angular, that on the long term it will make ngx-translate useless. So I think for the benefit of Squidex and its future, i should make a poc with Angular's i18n.

Then we can see the benefits of both and make a choice.

I tried to put that folder initially in the frontend, but I couldn't make the folder available for the 'server'? I knowledge about Webpack is not that big, but ill try it again.

Is there a way to make the nom make faster? the waiting times are a pain in the ass at this moment.

Summary:
I'll make a PoC with i18n so we can see the benefits of both choices. After that we/I can think about solutions for problems we face on that moment. 

I keep you updated.

SebastianStehle:
About npm: no, sorry....but once you have the initial setup done, you can keep npm start open, incremental builds are faster

MartijnDijkgraaf:
Is there a way to use the ng cli tool? because the `ng xi18n` function doesn't work because it needs a project definition. 

Otherwise I will look into the i18nOutFile options for the AngularCompilerPlugin in webpack, but right now I couldn't get any of this to generate the files for the translation...

MartijnDijkgraaf:
For me personally the ng i18n one is a no go...its hard to make it work and in the end you still have to deploy each language-specific version separately. So for me is ngx-translate the best choice.

SebastianStehle:
Makes sense as long as we abstract away the implementation, we can easily switch later.

MartijnDijkgraaf:
I have some time next week, so im going to make it as abstract as possible.

Ruud-cb:
Just want to share my 2 cents on translating. I've also used `ngx-translate`, in terms of tooling you could use Visual Studio and your C# project with the [ResXManager](https://marketplace.visualstudio.com/items?itemName=TomEnglert.ResXManager).

By using T4 files, you can generate the .json files easily. Since it's a relative directory from the backend that wouldn't be that hard to generate. Also gives you the possibility to use translatable 'things' on the backend as well if needed.
Have a look here: https://stackoverflow.com/questions/373388/best-way-to-implement-multi-language-globalization-in-large-net-project/35813707#35813707

MartijnDijkgraaf:
Hey, I'm back,

I had some things i needed to fix first, and my MacBook crashed. But when my workplace is set up I will look at this further. Where every form of help is appreciated

PR #533, opened 2020-06-09
Include Markdown editor and richtext editor in squidex instead of via CDN

Ruud-cb:
See ticket:
https://support.squidex.io/t/please-include-tinymce-js/2058 

This PR includes `simplemde` and `tinymce` within the project, to have no dependency on a CDN.  This is required for squidex instances running on-premise without internet. 

Code editor, Json editor and Geolocation editor components could also use these changes, this was for us beyond the scope of our requirements currently.

My first attempt was to include the `.js` files, but I couldn't get that to work with webpack. Thereafter I decided to install them via npm which is nicer anyway. There is however still one issue that I cannot seem to solve: I need to load the `.css` files as well, [importing them](https://github.com/Ruud-cb/squidex/commit/0151ad36c1942ce572bcf03929a8c6852b4726ee#diff-487b60ed923266b184065a3cae80642eR24) solves the visual issue but throws 404 console errors because the dependency is also trying to load those `.css` files. I am not sure how to fix this.

Never the less I hope this saves a bit of time, would love to see this PR being merged (or solved in another way) so that we can keep up with the latest changes when a new release is out.


SebastianStehle:
It makes everything too big, you have to find a solution to lazy load them.

Ruud-cb:
> It makes everything too big, you have to find a solution to lazy load them.

Would love to, but couldn't get webpack to include .js files in a separate "js" directory. If you can tell me how to make those changes then I can simply download the scripts from CDN, include them in frontend/js/xxx.js and change the path inside the components (url -> relative path).

SebastianStehle:
Perhaps: https://webpack.js.org/plugins/copy-webpack-plugin/ and then just use the resource loader again.

Ruud-cb:
Hi @SebastianStehle , I've updated as you have suggested, copy plugin seems to be the right way. However, when the code is trying to load the dependencies, it is locally searching at localhost:5001 instead of localhost:3000, When browsing to localhost:3000/dependencies/ace.js you will see the js file has been copied, just not loaded correctly in each component.
What solution to this would you suggest? Of course putting localhost:3000 in the `loadScript` as url is not an ideal solution (but does work). 
Logic inside each component to determine the URL? if environment is development use localhost url else a relative url? Perhaps there is already something available to this.

Ruud-cb:
Perhaps an unforeseen side effect, tinymce  requires to include quite a lot of plugin files...

Ruud-cb:
Ive added the suggestion of making the url depended on the environment, let me know how this can be improved.

SebastianStehle:
Then the only option is to put it in wwwroot folder of asp.net core or to make a differentiation between debug and production mode (what I prefer)

SebastianStehle:
I think we should also copy the dependencies from a npm package

Ruud-cb:
> I think we should also copy the dependencies from a npm package

I've included the dependencies of our dependencies that are required.
Not sure why it is not building right now in CI, I've already released this on-premise and it seems to work fine. Only thing that is unfortunately missing: FontAwesome. simplemde has a hardcoded dependency on a fontAwesome CDN.

SebastianStehle:
This can be used as environment variable: https://github.com/Squidex/squidex/blob/master/frontend/app/app.ts#L13

I think it would be better to add a dependency to the editors in npm and then copy and paste the necessary files from npm directly. This makes it much easier to upgrade to a new version.

SebastianStehle:
I was just thinking if it would not be easier to setup a configuration for this. Just a path to a location. Then you can setup it however you want, because I am very happy with the CDN solution and would not like to have a mix of several methods. Furthermore I am planing to integrate other editors as well.

Ruud-cb:
> This can be used as environment variable: https://github.com/Squidex/squidex/blob/master/frontend/app/app.ts#L13
> 
> I think it would be better to add a dependency to the editors in npm and then copy and paste the necessary files from npm directly. This makes it much easier to upgrade to a new version.

Sounds better yes, thought u didn't want them installed via npm, updated see last commits

> I was just thinking if it would not be easier to setup a configuration for this. Just a path to a location. Then you can setup it however you want, because I am very happy with the CDN solution and would not like to have a mix of several methods. Furthermore I am planing to integrate other editors as well.

I think the main question is: Do you want Squidex to support running on-premise and thus probably be more favourable for large organisations/governments which most likely don't allow internet connectivity (at least for internal apps)? 

I've done my part on avoiding a CDN to be used for some components, if you want to make this configurable then I will invite you to do so (after merge or in my fork repo). Unfortunately I don't have the time to do so and not enough info on how you want to do this, making it configurable while already copying dependencies from node_modules seems like a complex solution. Furthermore I don't see the point of a CDN in this particular case, sure it loads a bit faster for the first time, but after that it's cached anyways. And some of your CDN url's contained the "latest" version, seems undesired to us as this could result in an unexpected bug when a new version is pushed to the CDN. So you would need to lock on a version anyways, why not have that version maintainability inside your `project.json` by using npm? 

It's a bit unclear what the standard is, markdown editor for example uses font-awesome CDN, while the rich text editor has the icons inside it's package. Thus another change for on-premise is needed for us, which is include font-awesome as well. Unfortunately this results in loading font-awesome twice when you do have internet, a change (different branch) like [this is needed](https://github.com/Ruud-cb/squidex/commit/9fb4a6f9662f732dffe88dc679ed4be31b294f56) to support that. I can understand you don't want such a change in this PR.
 
Anyways, looking forward to your reply. As mentioned we need this anyway, so up until now we will use the [other branch](https://github.com/Ruud-cb/squidex/tree/release/4.3.x_modified) I have created (has the exact same changes with a little more like font-awesome). We hope this will at some point get merged to be more easily up-to-date with the latest version.

frontend/app-config/webpack.config.js
```diff
@@ -204,6 +206,26 @@ module.exports = function (env) {
                 // Add errors to webpack instead of warnings
                 failOnError: true
             }),
+            new plugins.CopyPlugin({
+                patterns: [
+                  { from: './node_modules/simplemde/dist', to: 'dependencies/simplemde' },
+
+                  { from: './node_modules/tinymce/tinymce.min.js', to: 'dependencies/tinymce/tinymce.min.js' },
```
SebastianStehle, line 213: I think I would use wildcard here and copy just everything.

frontend/app/framework/angular/forms/editors/code-editor.component.ts
```diff
@@ -82,8 +82,13 @@ export class CodeEditorComponent extends StatefulControlComponent<undefined, str
...
         }
-
-        this.resourceLoader.loadScript('https://cdnjs.cloudflare.com/ajax/libs/ace/1.2.6/ace.js').then(() => {
+        let url = 'dependencies/ace/ace.min.js';
+        if (process.env.NODE_ENV !== 'production') {
+            url = 'https://localhost:3000/' + url;
+        } else {
+            url = 'build/' + url;
+        }
+        this.resourceLoader.loadScript(url).then(() => {
```
SebastianStehle, line 91: Can we made this as apart of the resource loader? 

E.g. loadLocalScript and loadLocalStyle

And also use string interpolation please, I just always use it ;)

frontend/app-config/webpack.config.js
Ruud-cb: Are you sure? tinymce includes quite a lot of plugins and garbage files that are not needed. Will result in quite a lot of logs like so:

```
                                                                Asset        Size  Chunks             Chunk Names
                                               dependencies/ace/ace.js     371 KiB          [emitted]  
```

SebastianStehle: Makes sense.

SebastianStehle:
LGTM

SebastianStehle:
Can I merge it?

Ruud-cb:
One more question: Do you want to include font-awesome also? [These changes](https://github.com/Ruud-cb/squidex/commit/9fb4a6f9662f732dffe88dc679ed4be31b294f56)  would be applied to this branch (but then of course using `loadLocalStyle`)

SebastianStehle:
Otherwise it would look strange to have an editor without icons, right?

SebastianStehle:
Would you also make this change for the markdown editor and if possible other editors as well?

Ruud-cb:
Indeed, on-premise no icons visible in the markdown editor. Font awesome is added!
Looking at the usage of `loadScript`, I only see geolocation editor is still using this. Due it requires google maps and the many dependencies of leaflet, I would not want to invest time in this, perhaps at a later point.

The only thing that is now different from squidex version and our on-premise version is that we have removed codeblocks for outdated browser checks. Again because we cannot reach the CDN. If you see a solution for mitigating this then I am open for making modifications for this. But the path to such a local file included in the `index.html` is different from locally and production. I don't see any condition based solution inside the `index.html` file.
Can we by default hide this, and when `outdatedBrowser` function is present, unhide this?

SebastianStehle:
> Can we by default hide this, and when outdatedBrowser function is present, unhide this?

Yes, sounds like a good idea.

Ruud-cb:
I've actually fixed a bug of calling the `outdatedBrowser` function, it did not reach that code. Looking at the s[ource code ](https://github.com/outdatedbrowser/outdated-browser/blob/develop/outdatedbrowser/outdatedbrowser.js#L72), it will set `display: block` when needed. Tested by inserting random string inside `lowerThen` property and it correctly shows the warning.

In my opinion this can be merged now.

PR #536, opened 2020-06-17
Fix code editor, expected wrong ace.js file

Ruud-cb:
Hi Sebastian,

A small fix for the code editor, was expecting the wrong ace.js file. Apparently some caching in my browser didn't notice this. A `ace.min.js` also doesn't exist in node_modules.


SebastianStehle:
Thx

PR #550, opened 2020-07-27
I18n martijn

MartijnDijkgraaf:
This is the first setup for the frontend translations. (With location based replacing of the variables).

I still need to write tests and refactor the localizer.service.ts because I think that I can extract some functions.


frontend/app/shims.ts
```diff
@@ -5,6 +5,7 @@
  * Copyright (c) Squidex UG (haftungsbeschränkt). All rights reserved.
  */
 
+import '@angular/localize/init';
```
SebastianStehle, line 8: Why do we need this?

frontend/app/framework/utils/error.ts
```diff
@@ -15,6 +18,7 @@ export class ErrorDto {
         public readonly inner?: any
     ) {
         this.displayMessage = formatMessage(message, details);
+        this.localizer = LocalizerService.getInstance();
```
SebastianStehle, line 21: I think we should not use singleton here. I have only added it to the backend code because it was not possible otherwise. But I though we should just use the localizer in the components, when the error is displayed, not in errrodto

frontend/app/framework/services/title.service.ts
```diff
@@ -18,15 +19,15 @@ export class TitlesConfig {
...
+export const TitleServiceFactory = (titles: TitlesConfig, localizer: LocalizerService) => {
+    return new TitleService(titles, localizer);
 };
 
 @Injectable()
 export class TitleService {
     private readonly stack: any[] = [];
 
-    constructor(private readonly titles: TitlesConfig) {
+    constructor(private readonly titles: TitlesConfig, private readonly localizer: LocalizerService) {
```
SebastianStehle, line 30: Please use one line for each field that is directly assigned from the ctor. See other classes here.

frontend/app/framework/services/localizer.service.ts
```diff
@@ -0,0 +1,106 @@
...
+
+        if (args && args?.length > 0) {
+           text = this.replaceVariables(text, args);
+        }
+
+        return text;
+    }
+
+    private replaceVariables(text: string, args: readonly any[]): string {
+        const regex = new RegExp('\{(.*?)\}', 'g');
```
SebastianStehle, line 65: I think it would be good if we can solve this without regex. regex are really slow. Or at least a benchmark. Also only one regex is needed.

frontend/app/framework/services/localizer.service.ts
```diff
@@ -0,0 +1,106 @@
...
+            return arg;
+        }
+
+        switch (foundPipeOption[1]) {
+            case PipeOptions.TRANSLATE: {
+                return this.get(arg);
+            }
+
+            case PipeOptions.LOWER: {
+                return arg.charAt(0).toLowerCase() + arg.slice(1);
```
SebastianStehle, line 93: Slice works for strings? I was expecting substring here.

frontend/app/framework/services/localizer.service.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// TODO refactor this file, substract helper functions
+@Injectable()
+export class LocalizerService {
+    private static instance: LocalizerService;
+    public translations: Object;
+
+    constructor() {
+        if (process.env.NODE_ENV !== 'production') {
```
SebastianStehle, line 27: This should be part of the app.component not of the service itself, because it makes it impossible to test.

frontend/app/framework/services/dialog.service.ts
```diff
@@ -64,6 +65,7 @@ export class DialogService {
     private readonly requestStream$ = new Subject<DialogRequest>();
     private readonly notificationsStream$ = new Subject<Notification>();
     private readonly tooltipStream$ = new Subject<Tooltip>();
+    private readonly localizer = LocalizerService.getInstance();
```
SebastianStehle, line 68: No singleton please. Use dependency injection

frontend/app/framework/angular/http/http-extensions.ts
```diff
@@ -80,8 +80,10 @@ export module HTTP {
 
 export const pretifyError = (message: string) => <T>(source: Observable<T>) =>
     source.pipe(catchError((response: HttpErrorResponse) => {
-        const error = parseError(response, message);
+        const localizer = LocalizerService.getInstance();
```
SebastianStehle, line 83: See other comments below.

frontend/app/app.module.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import { CommonModule } from '@angular/common';
-import { HttpClientModule } from '@angular/common/http';
+import { HttpClientModule} from '@angular/common/http';
```
SebastianStehle, line 9: Please add the whitespace back.

frontend/app-config/webpack.config.js
```diff
@@ -329,7 +329,10 @@ module.exports = function (env) {
                 entryModule: 'app/app.module#AppModule',
                 skipCodeGeneration: !isAot,
                 sourceMap: !isProduction,
-                tsConfigPath: configFile
+                tsConfigPath: configFile,
```
SebastianStehle, line 332: What is all this? We do not use angular i18n

frontend/app/shims.ts
MartijnDijkgraaf: I will remove this.

frontend/app/app.module.ts
MartijnDijkgraaf: Done

frontend/app-config/webpack.config.js
MartijnDijkgraaf: Removed, I still had to clean it, my bad

frontend/app/framework/services/title.service.ts
MartijnDijkgraaf: Done

frontend/app/framework/services/localizer.service.ts
MartijnDijkgraaf: Slice and substring both work, but I can adjust it if you want.

frontend/app/framework/angular/http/http-extensions.ts
MartijnDijkgraaf: I dont see how we can do this with dependency injection, unless we edit the 157 uses of the pretityError.

Maybe you can help me out with this one?

frontend/app/framework/services/localizer.service.ts
MartijnDijkgraaf: I always tought that it depends on the regex. But if we can solve it without it a regex its has my preference.

I will think again about it.

MartijnDijkgraaf:
Thanks for the overall feedback!

frontend/app/framework/services/localizer.service.ts
SebastianStehle: You can have a look to C#.

frontend/app/framework/services/localizer.service.ts
SebastianStehle: I was just confused.

frontend/app/framework/angular/http/http-extensions.ts
SebastianStehle: Commented in the forum.

PR #555, opened 2020-08-16
Release/4.x martijn

MartijnDijkgraaf:


backend/i18n/translator/Squidex.Translator/Processes/GenerateKeys.cs
```diff
@@ -0,0 +1,39 @@
...
+        private readonly TranslationService service;
+        private readonly DirectoryInfo folder;
+
+        public GenerateKeys(DirectoryInfo folder, TranslationService service, string type)
+        {
+            if (type == "frontend")
+            {
+                this.folder = Frontend.GetFolder(folder);
+            }
+              else if (type == "backend")
```
SebastianStehle, line 26: There is a formatting error.

backend/i18n/translator/Squidex.Translator/State/TranslationService.cs
```diff
@@ -124,6 +124,18 @@ public void Save()
 
             Save("__todos.json", translationsTodo);
             Save("__ignore.json", translationToIgnore);
+            SaveKeys();
+        }
+
+        public void SaveKeys()
+        {
+            TranslationKeys translationKeys = new TranslationKeys();
```
SebastianStehle, line 132: Can you please use "var" and a new line?

backend/i18n/backend__keys.json
```diff
@@ -0,0 +1,325 @@
+{
```
SebastianStehle, line 1: Do we need this in the code? I mean it is only a temporary file that is used "nl", right?

MartijnDijkgraaf: Yea i can leave it out, if we need it we can generate it anyways.

backend/i18n/translator/Squidex.Translator/State/TranslationService.cs
MartijnDijkgraaf: I changed it to var, but what is the benefit of using var instead of a 'Type'?

SebastianStehle: Less code and consistency ;)

backend/i18n/translator/Squidex.Translator/Processes/GenerateKeys.cs
MartijnDijkgraaf: Fixec

backend/i18n/translator/Squidex.Translator/State/TranslationService.cs
```diff
@@ -124,6 +124,19 @@ public void Save()
 
             Save("__todos.json", translationsTodo);
             Save("__ignore.json", translationToIgnore);
+            SaveKeys();
```
SebastianStehle, line 127: I think this line does not make sense. We do not need a an empty key file as a permanent artifact

MartijnDijkgraaf: Yea your right, if somebody wants to add a new language they just need to run the command

PR #596, opened 2020-11-17
localization part 1

dborovsky:


SebastianStehle:
Hi, I have decided not to localize the script names, it is by purpose.

dborovsky:
> You should also read this: https://docs.squidex.io/01-getting-started/contributing-and-developing/translating
sure, i've already read this.

SebastianStehle:
But then you do not understand or the article is badly written. Because you should not change the translation files directly. Only change the files under source folder and then  run the translator.

dborovsky:
@SebastianStehle 
- To update localize files must I change corresponding files in folder backend/i18n/source
- Run the translate.bat or translate.sh

Correct?

SebastianStehle:
Yes, thats correct.

If you are not a native speaker, you should not translate these languages. There are other people and customers who take care about NL and IT

dborovsky:
> Hi, I have decided not to localize the script names, it is by purpose.

I returned back script without localization. 

I can't find any labels without localization. Could you show me this? 
Somewhere is no translation because localization key has value with english. Need translate to It and ND, but i am no native speaker.

SebastianStehle:
I think there is nothing to translate at the moment. Perhaps you misunderstood the original task.

PR #610, opened 2020-12-05
bug: fix identity server profile page showing client secret label as Client ID

alastairtree:
Fix this:

image


SebastianStehle:
Thank you very much :)

PR #612, opened 2020-12-12
Return contents with permission read.own

dborovsky:


backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
```diff
@@ -103,6 +103,13 @@ public async Task<IResultList<IEnrichedContentEntity>> QueryAsync(Context contex
                     contents = contents.SortSet(x => x.Id, q.Ids);
                 }
 
+                if (context.Permissions.Any(p => p.Id == Permissions.AppContentReadOwn))
+                {
+                    var claim = context.User.Claims.Where(p => p.Type == "sub").First().Value;
+                    var contentsList = contents.Where(c => c.CreatedBy.Identifier == claim).ToList();
+                    contents = ResultList.Create(contentsList.Count, contentsList);
```
SebastianStehle, line 110: This would filter the contents after they are received from MongoDB, we have to do it as part of the query...

SebastianStehle (changes requested):

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
dborovsky: To get  contents we use following query method: `var contents = await contentRepository.QueryAsync(context.App, schema, q, context.Scope());`

I thought to change params Q query.  Originally I wanted to do so.  You told me not to change  Filter. 

Can you show me example how to do it in query.

SebastianStehle: What I meant is not to add a parameter to q.ClrQuery.Filter, but you can add a property to Q.

dborovsky: Check my last comment. There is one issue. MongoDB.Driver for filter `filters.Add(Filter.Eq(x => x.CreatedBy.Identifier, query.CreatedBy));` displays error:  **Unable to determine the serialization information for x =\u003E x.CreatedBy.Identifier.**

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
```diff
@@ -92,6 +92,11 @@ public async Task<IResultList<IEnrichedContentEntity>> QueryAsync(Context contex
 
             var schema = await GetSchemaOrThrowAsync(context, schemaIdOrName);
 
+            if (context.Permissions.Any(p => p.Id == Permissions.AppContentReadOwn))
+            {
+                q.CreatedBy = context.User.Claims.Where(p => p.Type == "sub").First().Value;
```
SebastianStehle, line 97: There is an extension method for sub OpenIdSubject or so.

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
```diff
@@ -92,6 +92,11 @@ public async Task<IResultList<IEnrichedContentEntity>> QueryAsync(Context contex
 
             var schema = await GetSchemaOrThrowAsync(context, schemaIdOrName);
 
+            if (context.Permissions.Any(p => p.Id == Permissions.AppContentReadOwn))
```
SebastianStehle, line 95: Use could accidentally have other permissions that allow access to the content. Furthermore the permissions are not concrete, they contain placeholders. You have to use Permission.ForApp(<Permission>, appName, schemaName);

dborovsky: what should I pass in param **Id**?

SebastianStehle: Then permission

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
SebastianStehle: Ref token is just serialized to string. 

https://github.com/Squidex/squidex/blob/master/backend/src/Squidex.Infrastructure.MongoDb/MongoDb/RefTokenSerializer.cs

You should add a ref token to `Q` and create this with `q.CreatedBy = new RefToken(RefTokenType.User, user.OpenIdSubject())`

And then a simple filter:

 ```
filters.Add(Filter.Eq(x => x.CreatedBy, query.CreatedBy));
```

Comment (SebastianStehle) [reply backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs]: There is an extension method for user: https://github.com/Squidex/squidex/blob/835274396f58f9b9f8f11bab8956639337a97a5b/backend/src/Squidex.Infrastructure/Security/Extensions.cs#L16

Use this please.

**Comment by dborovsky** [comment]
@SebastianStehle check please changes in my last commit

File: backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs [hunk 4]
```diff
@@ -91,10 +92,14 @@ public async Task<IResultList<IEnrichedContentEntity>> QueryAsync(Context contex
             }
 
             var schema = await GetSchemaOrThrowAsync(context, schemaIdOrName);
+            var permissionReadOwn = Permissions.ForApp(Permissions.AppContentReadOwn, context.App.Name, schemaIdOrName);
 
-            if (context.Permissions.Any(p => p.Id == Permissions.AppContentReadOwn))
+            if (context.Permissions.Any(p => {
```
SebastianStehle, line 97: There is a method for that. You only need context.Permissions.Allows(permissionReadOwn)

but you also need to check if the user does not have the normal read permission.

backend/src/Squidex/Areas/Api/Controllers/Schemas/Models/SchemaDto.cs
```diff
@@ -103,6 +103,11 @@ protected virtual void CreateLinks(Resources resources)
                 AddGetLink("contents", resources.Url<ContentsController>(x => nameof(x.GetContents), values));
             }
 
+            if (resources.CanReadOwnContent(Name))
+            {
+                AddGetLink("contents", resources.Url<ContentsController>(x => nameof(x.GetContents), values));
```
SebastianStehle, line 108: The link name "contents" should be unique.

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
```diff
@@ -210,8 +208,13 @@ private async Task<List<ISchemaEntity>> GetSchemasAsync(Context context)
...
-            var permission = Permissions.ForApp(Permissions.AppContentsRead, context.App.Name, schema.SchemaDef.Name);
+            var permissionReadOwn = Permissions.ForApp(Permissions.AppContentReadOwn, context.App.Name, schema.SchemaDef.Name);
+            if (context.Permissions.Contains(permissionReadOwn))
+            {
+                return context.Permissions.Allows(permissionReadOwn);
+            }
 
+            var permission = Permissions.ForApp(Permissions.AppContentsRead, context.App.Name, schema.SchemaDef.Name);
             return context.Permissions.Allows(permission);
         }
```
SebastianStehle, line 219: The permission system is hierarchical:

`xxx.read` also includes `xxx.read.own`. So in this method we need to test if the user has permissions to read at least own contents. if this is the case we can make the query.

If the user does not have the permission to read all content, we must and the Own user id.

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
```diff
@@ -93,11 +93,9 @@ public async Task<IResultList<IEnrichedContentEntity>> QueryAsync(Context contex
 
             var schema = await GetSchemaOrThrowAsync(context, schemaIdOrName);
             var permissionReadOwn = Permissions.ForApp(Permissions.AppContentReadOwn, context.App.Name, schemaIdOrName);
```
SebastianStehle, line 95: Checking for permissionReadOwn  here is not necessary (See next comment).

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
dborovsky: Yes, I undestand the permission system is hierarchical.

If user doesn't have contents.read but has contents.read.own you mean to change something like this?

`var permissionReadOwn = Permissions.ForApp(Permissions.AppContentReadOwn, context.App.Name, schema.SchemaDef.Name);
            var permission = Permissions.ForApp(Permissions.AppContentsRead, context.App.Name, schema.SchemaDef.Name);

            if (!context.Permissions.Contains(permission))
            {
                return context.Permissions.Allows(permissionReadOwn);
            }
            
            return context.Permissions.Allows(permission);`

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
dborovsky: again.... If user doesn't have contents.read but has only contents.read.own  need I change to: 
```
if (context.Permissions.Allows(permissionRead))
            {
                q.CreatedBy = context.User.Token();
            }
```

SebastianStehle: Perhaps you should write some tests

backend/src/Squidex/Areas/Api/Controllers/Schemas/Models/SchemaDto.cs
dborovsky: If should be unique which name should I write for `resources.CanReadOwnContent(Name)`?

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryParser.cs
```diff
@@ -88,6 +88,11 @@ public virtual ValueTask<Q> ParseAsync(Context context, Q q, ISchemaEntity? sche
                 {
                     query.Take = options.DefaultPageSize;
                 }
+
+                if (q.CreatedBy != null)
```
SebastianStehle, line 92: I think this is not needed, looks wrong in my opinion.

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryService.cs
SebastianStehle: We only need to check for read.own here.

backend/src/Squidex.Infrastructure/Queries/Query.cs
```diff
@@ -24,6 +24,8 @@ public long Top
             set { Take = value; }
         }
 
+        public RefToken? CreatedBy { get; set; }
```
SebastianStehle, line 27: Please do not add it here.

backend/src/Squidex/Areas/Api/Controllers/Contents/Models/ContentDto.cs
```diff
@@ -184,6 +184,11 @@ private ContentDto CreateLinksAsync(IEnrichedContentEntity content, Resources re
                     AddPutLink("update", resources.Url<ContentsController>(x => nameof(x.PutContent), values));
                 }
 
+                if (resources.CanUpdateContentOwn(schema))
```
SebastianStehle, line 187: The resources is a dictionary, when you add 2 resources with the same name "update" you are overriding the previous value.

backend/src/Squidex/Areas/Api/Controllers/Schemas/Models/SchemaDto.cs
SebastianStehle: The resources is a dictionary, when you add 2 resources with the same name "update" you are overriding the previous value.

We only need to check for own permissions here in the API.

backend/src/Squidex.Infrastructure/Queries/Query.cs
dborovsky: I use it in QueryByQuery - CreateFilter

```
if (query?.CreatedBy != null)
            {
                filters.Add(Filter.Eq(x => x.CreatedBy, query.CreatedBy));
            }
```

How better than implement it?

backend/src/Squidex.Domain.Apps.Entities/Contents/Queries/ContentQueryParser.cs
dborovsky: If I delete this, user gets all contents.  But he must get only own contents.

backend/src/Squidex/Areas/Api/Controllers/Contents/Models/ContentDto.cs
dborovsky: Maybe something like this?

```
if (resources.CanUpdateContent(schema) || resources.CanUpdateContentOwn(schema))
                {
                    AddPutLink("update", resources.Url<ContentsController>(x => nameof(x.PutContent), values));
                }
```

SebastianStehle: You only need to check for .own

backend/src/Squidex.Infrastructure/Queries/Query.cs
SebastianStehle: I mean don't add it to Query, only to Q

dborovsky: If I remove this property, how should I rewrite this logic?

```
if (query?.CreatedBy != null)
            {
                filters.Add(Filter.Eq(x => x.CreatedBy, query.CreatedBy));
            }
```

SebastianStehle: Just take it from Q

dborovsky:
@SebastianStehle  check please last commit

backend/tests/Squidex.Domain.Apps.Entities.Tests/Contents/Queries/ContentQueryServiceTests.cs
```diff
@@ -218,6 +218,33 @@ public async Task QueryAll_should_not_return_contents_if_user_has_no_permission(
             Assert.Empty(result);
         }
 
+        [Theory]
+        [InlineData(1, 0, SearchScope.All)]
+        [InlineData(1, 1, SearchScope.All)]
+        [InlineData(0, 1, SearchScope.All)]
+        [InlineData(0, 0, SearchScope.Published)]
+        public async Task QueryAll_should_return_own_user_contents(int isFrontend, int unpublished, SearchScope scope)
```
SebastianStehle, line 226: I think what you want to test here is that q.Created is set depending on the user permissions. I don't see it in this test.

dborovsky: do you mean I need  test if permission is ReadOwnContent, query must to has filter CreatedBy?

backend/src/Squidex.Domain.Apps.Entities/Contents/DomainObject/Guards/GuardContent.cs
```diff
@@ -50,6 +48,11 @@ public static async Task CanCreate(CreateContent command, IContentWorkflow conte
         {
             Guard.NotNull(command, nameof(command));
 
+            if (!HasPermission(content, command, Permissions.AppContentsUpdate))
+            {
+                throw new DomainForbiddenException(T.Get("common.errorNoPermission"));
```
SebastianStehle, line 53: Please do not provide a message here, it is not used anyway.

backend/tests/Squidex.Domain.Apps.Entities.Tests/Contents/DomainObject/Guards/GuardContentTests.cs
```diff
@@ -247,6 +295,34 @@ public async Task CanChangeStatus_should_not_throw_exception_if_status_flow_vali
...
+            A.CallTo(() => contentWorkflow.CanMoveToAsync(content, content.Status, command.Status, userWithStatusChangePermission))
+                .Returns(true);
+
+            await GuardContent.CanChangeStatus(command, content, contentWorkflow, contentRepository, schema);
+        }
+
+        [Fact]
+        public async Task CanChangeOwnStatus_should_throw_exception_if_has_no_permission()
+        {
+            var schema = CreateSchema(false);
```
SebastianStehle, line 315: Avoid code duplication in tests.

backend/src/Squidex.Domain.Apps.Entities/Contents/DomainObject/Guards/GuardContent.cs
dborovsky: but DomainForbiddenException must to have params string

SebastianStehle: Thats true, but then remember to translate it as well.

dborovsky: But I'll get error. I must to pass any string there. What you mean **do not provide a message here, it is not used anyway**

backend/tests/Squidex.Domain.Apps.Entities.Tests/Contents/DomainObject/Guards/GuardContentTests.cs
dborovsky: I can refToken to move to  test class as property and use anywhere like this:
`private readonly RefToken refToken = new RefToken("string", "string");`

Do you commands, contents too?

SebastianStehle: This does not make sense, I mean you could just create a helper method or so, but creating one field that is only needed for one test just moves the duplication from the method to the class level.

PR #614, opened 2021-01-05
fix search.help url

alexvolchetsky:
fix incorrect url that is used here:
image


SebastianStehle:
Thank you very much. Please read this page first: https://docs.squidex.io/01-getting-started/contributing-and-developing/translating

As a summary: you also have to change everything under i18n/source

alexvolchetsky:
thanks. updated files in i18n/source also

PR #616, opened 2021-01-08
Feat/enhanced oidc settings

alastairtree:
Hello

This PR contains a bugfix and some minor enhanced config/support for the external OIDC feature. It has been designed so that it maintains backwards compatibility and does not change the existing external oidc settings, but allows for some additional config. I have tested it with my oidc server and it works well now, whereas the current config options do not work as our server is a bit quirky.

First the bug.
The recent changes to IdentityServerServices for local config, replacing config in `authBuilder.AddOpenIdConnect` and the line below has an undesirable side-effect - it removes the default IPostConfigureOptions<OpenIdConnectOptions> event meaning that the external oidc config has stopped working because it has not been set up correctly and instead throw errors. I have restored the use of AddOpenIdConnect but with a small hack to acquire the DI container by building it. It is not as elegant but it does seem to work. You may wish to concoct another way of getting the service provider though. Either way it needs fixing - all use of external oidc is currently broken.
```
 authBuilder.Services.AddSingleton<IPostConfigureOptions<OpenIdConnectOptions>>(c => new PostConfigureOptions<OpenIdConnectOptions>(OpenIdConnectDefaults.AuthenticationScheme, options =>
```

I have also added support for external oidc providers who put the email in a claim "email" rather than the existing claim value of ClaimTypes.Email which is "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/email", but have done so as a fallback leaving the current behaviour in place. Otherwise the empty claim generates an exception.

Secondly the new OIDC config.
I have added new oidc variables
* oidcMetadataAddress - enabling OIDC auto discovery. Default to blank/disabled
* oidcResponseType - leave the current id_token default, but allow the user to override it with other options such as "code" required for some OIDC servers
* oidcGetClaimsFromUserInfoEndpoint - tell aspnet to get claims from the oidc server on the backchannel rather than read them from the JWT. existing default of false is maintained
* oidcOnSignoutRedirectUrl - allow you to override the default logout behaviour of redirecting to your oidc server logout url if that is not supported/desirable and specifiy an alternative such as "/". Leaving it blank as the default is means normal oidc logout route is followed.

Please let me know any issues/work required.

Thanks for a great product!

Alastair

### DIFF
--- a/backend/src/Squidex.Infrastructure/Security/Extensions.cs
+++ b/backend/src/Squidex.Infrastructure/Security/Extensions.cs
@@ -67,6 +67,12 @@ namespace Squidex.Infrastructure.Security
             return principal.Claims.FirstOrDefault(x => x.Type == OpenIdClaims.Email)?.Value;
         }
 
+        public static string? TryFindEmail(this ClaimsPrincipal principal)
+        {
+            return principal.Claims.FirstOrDefault(x => x.Type == ClaimTypes.Email)?.Value ??
+                   principal.Claims.FirstOrDefault(x => x.Type == OpenIdClaims.Email)?.Value;
+        }
+
         public static bool IsInClient(this ClaimsPrincipal principal, string client)
         {
             return principal.Claims.Any(x => x.Type == OpenIdClaims.ClientId && string.Equals(x.Value, client, StringComparison.OrdinalIgnoreCase));
--- a/backend/src/Squidex/Config/Authentication/OidcHandler.cs
+++ b/backend/src/Squidex/Config/Authentication/OidcHandler.cs
@@ -40,5 +40,19 @@ namespace Squidex.Config.Authentication
 
             return base.TokenValidated(context);
         }
+
+        public override Task RedirectToIdentityProviderForSignOut(RedirectContext context)
+        {
+            if (!string.IsNullOrEmpty(options.OidcOnSignoutRedirectUrl))
+            {
+                var logoutUri = options.OidcOnSignoutRedirectUrl;
+                context.Response.Redirect(logoutUri);
+                context.HandleResponse();
+
+                return Task.CompletedTask;
+            }
+
+            return base.RedirectToIdentityProviderForSignOut(context);
+        }
     }
 }
--- a/backend/src/Squidex/Areas/IdentityServer/Controllers/Account/AccountController.cs
+++ b/backend/src/Squidex/Areas/IdentityServer/Controllers/Account/AccountController.cs
@@ -294,7 +294,7 @@ namespace Squidex.Areas.IdentityServer.Controllers.Account
             }
             else
             {
-                var email = externalLogin.Principal.FindFirst(ClaimTypes.Email)?.Value!;
+                var email = externalLogin.Principal.TryFindEmail();
 
                 user = await userManager.FindByEmailWithClaimsAsync(email);
 
--- a/backend/src/Squidex/Areas/IdentityServer/Controllers/Extensions.cs
+++ b/backend/src/Squidex/Areas/IdentityServer/Controllers/Extensions.cs
@@ -12,6 +12,7 @@ using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication.OpenIdConnect;
 using Microsoft.AspNetCore.Identity;
+using Squidex.Infrastructure.Security;
 using Squidex.Web;
 
 namespace Squidex.Areas.IdentityServer.Controllers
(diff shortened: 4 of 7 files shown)


SebastianStehle:
Thank you very much. I don't understand why it removes the default IPostConfigureOptions. I was debugging it and found out that multiple PostConfigureOptions seem to running (4 exactly).

alastairtree:
Because you have added it as a singleton I think it overrides and replaces the other oidc options config

backend/src/Squidex.Infrastructure/Security/Extensions.cs
```diff
@@ -67,6 +67,12 @@ public static class Extensions
             return principal.Claims.FirstOrDefault(x => x.Type == OpenIdClaims.Email)?.Value;
         }
 
+        public static string? TryFindEmail(this ClaimsPrincipal principal)
```
SebastianStehle, line 70: Can you not do this in the OpenIdHandler? In TokenValidated method...

backend/src/Squidex/Config/Authentication/IdentityServerServices.cs
```diff
@@ -50,15 +51,17 @@ public static AuthenticationBuilder AddSquidexIdentityServerAuthentication(this
...
                 if (!string.IsNullOrWhiteSpace(identityOptions.AuthorityUrl))
                 {
                     options.Authority = identityOptions.AuthorityUrl;
                 }
                 else
                 {
-                    options.Authority = GetAuthorityUrl(c);
+                    // probably a better way to acquire DI but this works
+                    var provider = authBuilder.Services.BuildServiceProvider();
+                    options.Authority = GetAuthorityUrl(provider);
```
SebastianStehle, line 64: I don't like this, it is really a hack, we should try to understand the options system better.

SebastianStehle:
When you debug through you see multiple post configures

image

backend/src/Squidex/Config/Authentication/OidcHandler.cs
```diff
@@ -40,5 +40,19 @@ public override Task TokenValidated(TokenValidatedContext context)
 
             return base.TokenValidated(context);
         }
+
+        public override Task RedirectToIdentityProviderForSignOut(RedirectContext context)
+        {
+            if (!string.IsNullOrEmpty(options.OidcOnSignoutRedirectUrl))
+            {
+                var logoutUri = options.OidcOnSignoutRedirectUrl;
```
SebastianStehle, line 48: The OpenIdOptions has SignedOutRedirectUri...what is wrong with that?

SebastianStehle (changes requested):

SebastianStehle:
I have tested another approach, which should be hopefully correct.

alastairtree:
Had a look at the branch and looks like what I was going to try out so looks promising. Will take a look at your comments and try out out on Monday. Thanks.

backend/src/Squidex/Config/Authentication/OidcHandler.cs
alastairtree: Because it is not the same this. `SignedOutRedirectUri` is the location to send browser after all the logout is done. 

If your External OICD provider does not support logout then the logout url returned during metatdata discovery from oidcMetadataAddress is empty. When the external redirect URL is empty the following exception is thrown even if you specify SignedOutRedirectUri

`InvalidOperationException: Cannot redirect to the end session endpoint, the configuration may be missing or invalid.`

More info in https://stackoverflow.com/questions/49815961/cannot-redirect-to-the-end-session-endpoint-the-configuration-may-be-missing-or

So to fix you need to handle the logout event yourself - hence allowing config to override it here.

backend/src/Squidex.Infrastructure/Security/Extensions.cs
alastairtree: I checked and if you are using `options.GetClaimsFromUserInfoEndpoint` then the claims are not present in TokenValidated - they are fetched afterwards, so you need to handle it when searching the claims.

backend/src/Squidex/Config/Authentication/IdentityServerServices.cs
alastairtree: You new code works, and is MUCH nicer. Will rebase,

alastairtree:
@SebastianStehle I have rebased, answered your queries, and tested it locally with our external oidc provider and it is working nicely. Let me know if you need anything else.
Thanks.

SebastianStehle:
Awesome, thanks a lot.

alastairtree:
Thank you for merging @SebastianStehle. Are you able to give any indication of when this might appear in a docker release?

PR #657, opened 2021-02-12
fix API description for docs

EdoardoTona:

PR #661, opened 2021-02-28
Wrong returned type AssetFoldersDto

EdoardoTona:
This fix an issue with the folders methods: they have wrong documentation about the returned type: it's declared `AssetsDto` but they are returning `AssetFoldersDto`. This produce wrong generated code (e.g. in the C# SDK).

After the merging the C# SDK generated code should be updated.

https://github.com/Squidex/squidex/pull/657 is proposed again (it produces a wrong docs description, and create confusion with another api with the same name)


SebastianStehle:
Thank you very much, I have already fixed this in another branch.

PR #683, opened 2021-04-13
Revert back to agnostic tag

richlander:
- `.102-ca-patch-buster-slim` is no longer serviced
- The latest `5.0` patch has related fixes.

Context: https://github.com/dotnet/announcements/issues/180

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 #
 # Stage 1, Build Backend
 #
-FROM mcr.microsoft.com/dotnet/sdk:5.0.102-ca-patch-buster-slim as backend
+FROM mcr.microsoft.com/dotnet/sdk:5.0 as backend
 
 ARG SQUIDEX__VERSION=4.0.0
 


richlander:
As an aside, do you have any feedback on how .NET container images can be improved? Do they meet your expectations?

SebastianStehle:
They work pretty well, but in case of error handling and debugging they are very limited.

There are a few tools that would be helpful for these cases:

* curl
* ping
* remote debugger
* dotnet dump

I understand that you want to keep them as small as possible, but perhaps there is a solution for that. Perhaps a script to install the debug tools inside the container or something like this. It took me a while to find all these information and when you have issues with your production system this is not what you want to do.

richlander:
Have you considered installing those yourself in your Dockerfile, like `curl` for example?

shirhatti:
@SebastianStehle Happy to chat more especially about the remote debugger (vsdbg) and dotnet-dump use case.

In general, our guidance is to use multi-stage builds to add exactly whatever tools you need in addition to whatever's baked into the dotnet base images.

In situations where an issue has already occurred and you don't have the diagnostics tools already baked in, you can acquire a single-file version of the diagnostics tool. 

For example, if you’re running .NET Core on x64 Ubuntu, you can acquire the dotnet-dump from https://aka.ms/dotnet-dump/linux-x64.

SebastianStehle:
@richlander Sure, but when I realized that I was in the process to investigate the issue. 

@shirhatti Thats great, will test them out next time. Yesterday I had to create a dump and I had to install the SDK on the container and then dotnet-tools.

richlander:
Makes sense @SebastianStehle. Part of what you are experiencing is just the nature of containers. For example, we've had people asking us to support [distroless](https://github.com/GoogleContainerTools/distroless). Those images don't even have a shell.

Main point is that -- for any development stack -- observability and diagnosability requires upfront design decision and effort to work well, both for containers and generally. That's, in part, why we built [dotnet-monitor](https://devblogs.microsoft.com/dotnet/introducing-dotnet-monitor/) and support multiple deployment approaches to align with the different needs people have.

PR #697, opened 2021-05-13
AssetRequestSizeLimit not configured for asset replacement

Jaben:
@SebastianStehle -- the upload limit is not set for asset replacements -- I believe this is the fix.

Thanks for all your effort on Squidex -- it's a fantastic app.


SebastianStehle:
Thx you very much :)

Issue #692, opened 2021-05-05
Bring Squidex to marketplaces


SebastianStehle:
Bring Squidex to the different marketplaces:

* [x] Vultr Marketplace (https://www.vultr.com/marketplace/apps/squidex)
* [ ] Digital Ocean App
* [x] Digital Ocean Droplet
* [ ] Digital Ocean Kubernetes
* [ ] Render
* [ ] Heroku (not sure if possible)
* [ ] Azure
* [x] AWS
* [ ] GCP

PR #704, opened 2021-05-17
Add GitHub Actions Workflows

kamushadenes:
# Description

This pull request migrates the existing workflow in `.drone.yml` to GitHub Actions, and also creates some placeholder files to publish to different marketplaces as per #692, which will be done in future pull requests.

# Requirements

The following secrets must be created in the repository:

- DOCKER_USERNAME
- DOCKER_PASSWORD

# Next Steps

If approved, the `.drone.yml` file should be removed to avoid duplicating the work.


.github/workflows/dev.yml
```diff
@@ -0,0 +1,96 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Inject slug/short variables
+        uses: rlespinasse/github-slug-action@v3.x
+      - name: Replace Image Name
+        uses: mikefarah/yq@v4.9.1
+        with:
+          cmd: yq e '.services.squidex.image = "squidex/squidex:tmp-dev-${{ env.GITHUB_SHA_SHORT }}"' -i backend/tests/docker-compose.yml
```
SebastianStehle, line 51: Can we just simple increment version as we have it now. I use this approach in another project:

https://github.com/notifo-io/notifo/blob/main/.github/workflows/deploy.yml#L16

.github/workflows/dev.yml
```diff
@@ -0,0 +1,96 @@
...
+          build-args: "SQUIDEX__VERSION=4.0.0-dev-${{ env.GITHUB_SHA_SHORT }}"
+          cache-from: type=registry,ref=squidex/squidex:cache-dev
+          cache-to: type=registry,ref=squidex/squidex:cache-dev
+
+  test:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
SebastianStehle, line 45: Can you add an empty line between steps. It is very hard to scan like this.

.github/workflows/dev.yml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build and Push Temporary Tag
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: squidex/squidex:tmp-dev-${{ env.GITHUB_SHA_SHORT }}
+          build-args: "SQUIDEX__VERSION=4.0.0-dev-${{ env.GITHUB_SHA_SHORT }}"
+          cache-from: type=registry,ref=squidex/squidex:cache-dev
```
SebastianStehle, line 37: The problem with the cache seems to be that it grows forever right now. I had to this:

https://github.com/notifo-io/notifo/blob/main/.github/workflows/deploy.yml#L28

SebastianStehle (changes requested):

.github/workflows/dev.yml
kamushadenes: According to [cache.md](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md), this problem only happens when using actions/cache. This method uses the registry, so there should be no issue.

.github/workflows/dev.yml
kamushadenes: Sure!

.github/workflows/dev.yml
kamushadenes: Sure!

.github/workflows/dev.yml
SebastianStehle: For both approaches it is unclear for me what happens when you run it in the background.

.github/workflows/dev.yml
SebastianStehle: Please use the trick with the increment, I don't want that dev starts from 1 again.

.github/workflows/dev.yml
kamushadenes: It prevents redoing some steps.

Honestly, as the Dockerfile is now, a clean setup would only benefit on Stage 3 on the apt-get run, as any COPY invalidates all upper layer caches. 

Since almost all stages start with a COPY command, the benefit is minimal. Would you like to remove?

.github/workflows/dev.yml
kamushadenes: Done

.github/workflows/dev.yml
SebastianStehle: No, it is beneficial:

1. Because i copy the pacakges.json and csproj first i can very often skip npm install and dotnet restore
2. Because I have 2 separate steps for frontend and backend I can skip the frontend if I only change the backend.

Honestly it does not work reliable, but it should be fine.

.github/workflows/dev.yml
```diff
@@ -0,0 +1,128 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and Push Temporary Tag
```
SebastianStehle, line 42: Just another request. Can you write the major step in uppercase, because I quickly need to jump to this step very often.

.github/workflows/dev.yml
```diff
@@ -0,0 +1,128 @@
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
```
SebastianStehle, line 39: We cannot use secrets for pull requests from forks.

.github/workflows/dev.yml
```diff
@@ -0,0 +1,128 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: BUILD AND PUSH TEMPORARY TAG
+        uses: docker/build-push-action@v2
+        with:
+          push: true
```
SebastianStehle, line 45: We should not push for pull-request

.github/workflows/dev.yml
```diff
@@ -0,0 +1,128 @@
...
+          tags: squidex/squidex:tmp-dev-${{ env.BUILD_NUMBER }}
+          build-args: "SQUIDEX__VERSION=4.0.0-dev-${{ env.BUILD_NUMBER }}"
+          cache-from: type=registry,ref=squidex/squidex:cache-dev
```
SebastianStehle, line 48: Might not work, because we cannot use secrets for PR.s

PR #711, opened 2021-05-27
reordering services configuration

sebestyn168:
Hello,

Using plugin for application template fails when executing the creation commands

During HandleAsync, context.IsCompleted is false and prevents the creation to be handled
```
public async Task HandleAsync(CommandContext context, NextDelegate next)
{
      if (context.IsCompleted && context.Command is CreateApp createApp && IsRightTemplate(createApp))

```

To be specific, it seems that `services.AddSquidexMvcWithPlugins(config);` needs to be called after 
```
            services.AddSingletonAs<SingletonCommandMiddleware>()
                .As<ICommandMiddleware>();
```
in CommandsServices.cs, thus we moved `AddSquidexMvcWithPlugins` after `AddSquidexCommands`

We're not fully aware of potential side effect this request could have. 
It works for our template applications plugins.

Regards

### DIFF
--- a/backend/src/Squidex/Startup.cs
+++ b/backend/src/Squidex/Startup.cs
@@ -38,14 +38,13 @@ namespace Squidex
             services.AddMemoryCache();
             services.AddNonBreakingSameSiteCookies();
 
-            services.AddSquidexMvcWithPlugins(config);
-
             services.AddSquidexApps();
             services.AddSquidexAssetInfrastructure(config);
             services.AddSquidexAssets(config);
             services.AddSquidexAuthentication(config);
             services.AddSquidexBackups();
             services.AddSquidexCommands(config);
+            services.AddSquidexMvcWithPlugins(config);
             services.AddSquidexComments();
             services.AddSquidexContents(config);
             services.AddSquidexControllerServices(config);


SebastianStehle:
I think there is a ICustomCommandMiddleware interface to solve this problem. I don't remember the reason but there is one why the order is like that (I know, comments would help)

sebestyn168:
I tried to inherit our class from `ICustomCommandMiddleware`
`public sealed class CreateTestClassCommandMiddleware : ICustomCommandMiddleware`
and update the command registration
```
            services.AddSingletonAs<CreateTestClassCommandMiddleware>()
               .As<ICustomCommandMiddleware>();
```
but creating the app ends up with a blank app with no schema
Even while keeping `services.AddSquidexMvcWithPlugins(config);` _after_ `services.AddSquidexCommands(config);`

with a break point on our `HandleAsync` method
`if (context.IsCompleted && context.Command is CreateApp createApp && IsRightTemplate(createApp))`
the first time its reached `context.Command` is a `CreateApp` object but `context.IsCompleted` is `false`

SebastianStehle:
How do you call next() in your middleware?

SebastianStehle:
I have pushed a sample for your use case.

sebestyn168:
It works great, thank you!

SebastianStehle:
have you seen the new interface?

sebestyn168:
ITemplate ?
Yes I inherited from it in our templates.
It removes the command Handling process so implementation is straight forward now.
So great job as it's more developer friendly.

PR #715, opened 2021-05-30
add multiple providers

### LINKED ISSUE #692: Bring Squidex to marketplaces

kamushadenes:
This PR introduces multiple providers:

- [x] Azure
- [x] GCP
- [x] Vultr
- [x] Heroku
- [x] Render.com

It also adds a new section to the README.md with the deploy button for Heroku and Render.com. Once images get listed to the respective marketplaces, we can add buttons for other platforms as well.

There are some secrets to be configured and some non-trivial tasks to be performed on Azure and GCP before this PR can be merged. 

After this PR, only Helm is left to close #692


SebastianStehle:
Hi, great work so far. I have pushed it so that I could test heroku and render.com and this is my conclusion so far:

## Render.com

1. It definitely needs doku, because you cannot describe the env vars. There should be doku for default domain and custom domain.
2. Why do we need EVENTSTORE__TYPE? 
3. If no domain is entered, can we get the default domain?
4. If no GOOGLE_CLIENT_ID is entered, the env variable will have the default from dev environment. We need to configure it with an empty string instead to get rid of this login option.
5. Why is there a wait 60? It is rendered in the logs and not very descriptive.
6. Is scaling possible?
7. I don't understand when the server restarts when changing env variables.
7. The generated password does not work: 

```
  "exception": {
May 31 01:56:49 PM      "type": "Squidex.Infrastructure.Validation.ValidationException",
May 31 01:56:49 PM      "message": "Passwords must have at least one non alphanumeric character.",
```

## Heroku

1. The description for the env vars could be better
2. We should now show the XXX_CLIENT_ID things as it makes the setup more complicated.

kamushadenes:
Hey Sebastian,

## Render.com

1. Documentation is in the works
2. I don't know, I copied the environment vars from https://github.com/Squidex/squidex-hosting/blob/master/docker-compose/docker-compose.yml
3. We can do that in the script, yes. What is the default domain?
4. Will do
5. Application was started before MongoDB could finish initializing and there was no auto restart, so I had to add it.
6. Yes, after the deployment, the user can simply scale the webserver from the UI
7. This is a render.com behavior, I'm not sure we can change it
8. I'll remove it then since I can't customize the generator
9. I'll add `autoDeploy: false`

## Heroku

1. Documentation is in the works
2. I'm not familiar with how the application works internally, could you please make some suggestions?
3. I'll remove them. Should do the same for Render.com?
4. I'll remove it
5. Can you elaborate? That seems to depend on the application.

## General
- I'll investigate
- Could you please elaborate?
- It's a managed image. During the marketplace setup, we need to investigate the need of a Shared Gallery Image or some other approach.

SebastianStehle:
Hey Sebastian,

## Render.com

1. Documentation is in the works
2. Just remove it.
3. It is something from render.com, e.g. in heroku it is [project-name].heroku.com or so
4. I see
5. I see
6. What I mean: The instances need to talk to each other with port 40.000
7.  It seems that it restarts automatically. We should add it to the docs.
8. It is easier to create the user with the setup page anyway.
9. Great

## Heroku

1. Documentation is in the works
2. I'm not familiar with how the application works internally, could you please make some suggestions?
3. I'll remove them. Should do the same for Render.com?
4. I'll remove it
5. For Squidex, clustering to work needs to talk over port 40.000. I am not sure, if it is possible with heroku

PR #718, opened 2021-06-04
Small Fixes for Heroku and Render.com

kamushadenes:
Please see https://github.com/Squidex/squidex/pull/715#issuecomment-851443770


app.json
```diff
@@ -18,41 +18,9 @@
       "description": "Mongo Password (see MONGO_USERNAME)",
       "value": "Squidex123"
     },
-    "EVENTSTORE__TYPE": {
-      "description": "EventStore Type",
-      "value": "MongoDB"
-    },
     "IDENTITY__ADMINEMAIL": {
```
SebastianStehle, line 21: You can also remove the admin email because the initial admin can be configured with a setup screen.

packer/heroku/squidex/squidex.sh
```diff
@@ -10,4 +10,6 @@ export STORE__MONGODB__CONTENTDATABASE="Squidex"
 
 export ASPNETCORE_URLS="http://+:$PORT"
 
+export IDENTITY__GOOGLECLIENT=""
+
```
SebastianStehle, line 14: I think we have to do it as well for all other cloud providers like GITHUB and MICROSOFT

packer/render/squidex/squidex.sh
```diff
@@ -1,10 +1,16 @@
...
 
+if [ "x$DOMAIN" = "x" ]; then
+    export DOMAIN=$RENDER_EXTERNAL_HOSTNAME
+fi
+
 export URLS__BASEURL="https://${DOMAIN}"
 export EVENTSTORE__MONGODB__CONFIGURATION="mongodb://${MONGO}"
 export STORE__MONGODB__CONFIGURATION="mongodb://${MONGO}"
 
+export IDENTITY__GOOGLECLIENT=""
```
SebastianStehle, line 14: I think we have to do it as well for all other cloud providers like GITHUB and MICROSOFT

render.yaml
```diff
@@ -18,24 +19,8 @@ services:
           name: squidex_mongo
           type: pserv
           property: hostport
-      - key: EVENTSTORE__TYPE
-        value: MongoDB
       - key: IDENTITY__ADMINEMAIL
```
SebastianStehle, line 22: You can also remove the admin email because the initial admin can be configured with a setup screen.

SebastianStehle (changes requested):
See my comments

kamushadenes:
Done

PR #722, opened 2021-06-05
Add backend_zh.json

seamys:
Add backend_zh.json (zh_CN)


SebastianStehle:
I will make the necessary change tomorrow so that you can test it yourself.

seamys:
The `frontend_en.json` has too many items that need to translate, I need more time to do it.

SebastianStehle:
Yes, of course. But I need to make the changes to enable chinese language. It will fallback to English then and you can easily test it.

PR #736, opened 2021-07-15
Release/design2

SebastianStehle:


MartijnDijkgraaf:
**Tested on: Dev-6150**

Overall comment:
Try to make use of cssGrid, it makes it easy to create template layouts.

- The Icons in front of after texts are weird aligned.
For example: 
image
image
- On the assets page: the taskbar is aligned in the middle. Therefore you cant read the placeholder of "Search by name". This also occures on the 'administration/users' page. 
image 
- On safari some buttons look weird (when i remove the -webkit-appearance: button;) its fixed.
image
- On the rules page: there is top right an angle-right visible, but this doesn't seem to do anything (I think this should became visible when you open the history panel.
- The panels on the Schema's and Content page arent the same width (the add button makes the Schemas panel bigger)
-  On the edit content editor mode, there is a chrevron left visible. (After the Tab title)
image
- On the edit content editor mode, the top right buttons are good aligned 
image
image
- On the edit content editor mode, I prefer the buttons to be aligned with the contents
image
Something like this: (but then better)
image
- On the edit content page: When the panel is open, the content sticks to the panel. 
image
I prefer this:
image
- On contents content overview page, Checkbox in front of action is not right aligned
image
- Why is the word status been truncated, there is enough space
image

PR #737, opened 2021-07-16
minor adjustment to correct en translation

richorama:
Hi Sebastian,

I know this is a really minor typo, but the English translation is not quite correct, and it's one of the first things you see in the app.

:¬)


SebastianStehle:
Thank you very much :) ... You are right. the first impression matters :)

PR #739, opened 2021-07-23
Fix issue with tag retrieval

pavliy:
- Align schemaproperties tags property type for correct map later in pipe

Just to be more specific. If now user tries to save "Schema" tags - saved tags are not returned and tags field is always empty. That's because SchemaProperties uses custom ImmutableList, while SchemaPropertiesDto - uses standard .NET immutablelist.
This model alignment will allow SimpleMapper to convert tags correctly


SebastianStehle:
Thanks a lot

PR #749, opened 2021-08-27
Reposition labels orleans

sanderverhoeven-weareyou:
repositioning of the labels:
        orleans/serviceId: "squidex-{{ default .Chart.AppVersion .Values.image.tag }}"
        orleans/clusterId: "squidex"


SebastianStehle:
I have never used helm before but I think adding the image tag to the service id is not a good idea. You would basically spin up a new cluster with each tag and then during the deployment you would have multiple grain instances of central services like the scheduler.

sanderverhoeven-weareyou:
@SebastianStehle I was on leave last week. I only repositioned the label. So you mean this ID has to be the same every time?

SebastianStehle:
You mean you have solved the error?

sanderverhoeven-weareyou:
Yes just solved the error when deploying with helm no other changes were made beside the package afcourse.

sanderverhoeven-weareyou:
Now we are talking could you add the following to the K8s Documentation for adding the repository:

helm repo add squidex https://squidex.github.io/squidex/helm
helm repo update

on page : https://docs.squidex.io/01-getting-started/installation/platforms/install-on-kubernetes

SebastianStehle:
You can also do it yourself: https://github.com/squidex/squidex-docs2 ;)

Thank you very much

PR #758, opened 2021-09-07
Removed invalid markup from links

slalFe:
Previously clicking on these links would take you to 'Page not found', now they take you to the home page as expected.


SebastianStehle:
Thanks a lot.

PR #775, opened 2021-10-27
WIP | Language dropdown enhancement

pjmagee:
Feature Enhancement: https://support.squidex.io/t/need-help-schema-filter-language-dropdown-changes/3540

The PR code has only been updated to fix compile errors, but it seems the logic is now broken with the latest version of Squidex. I am not sure exactly where in the logic that it needs to change to work like before. But this PR does give you a  preview of what the implementation looks like.

Please review / comment etc. 

image

image

image

image


pjmagee:
> I just had a look and checked the code to see what i can find.

I fixed it. Now, after reviewing it again. I have some thoughts.

- I don't feel comfortable or confident enough to move this into some kind of state file, not sure how i would do it and im not very good. I was thinking maybe this could be added to the languages.state file, however the computing needs to happen on each paging of the contents, so im not sure again how that would best be done.

- It's relatvily simple and isolated as a feature in its current form, and just hooks into the current schemas and contents in state, and then attaches a Map of language boolean flags to identify which content has language data set or not. So maybe it could be added like this and then improved in the future.

SebastianStehle:
But there are still other issues. Every field is defined when not empty or not null. There is no reason to handle arrays differently

pjmagee:
> But there are still other issues. Every field is defined when not empty or not null. There is no reason to handle arrays differently

Maybe I misunderstand something then, originally I was not checking if something was an array and then I was getting different results than I expected. But after handling it like this, it was working well with the schemas I have setup. I'll talk to Simon and see what I can do to simplify it.

SebastianStehle:
The definition of valid is this one: https://github.com/Squidex/squidex/blob/master/frontend/app/framework/angular/forms/forms-helper.ts#L158

You could export it and give it a better name.

SebastianStehle:
Btw: I think you have forgotten to update the language selector in more places:

* content-page component 
* content-selector
* content-creator

We should either make it only on field level or on all levels to be consistent.

pjmagee:
> Btw: I think you have forgotten to update the language selector in more places:
> 
> * content-page component
> * content-selector
> * content-creator
> 
> We should either make it only on field level or on all levels to be consistent.

Okay, I agree. I will take a look at this tomorrow.

SebastianStehle:
I am not sure if you have seen my comments, but these 3 things are important:

1. Because so many components need this feature it is important to extract the methods to a separate file.
2. The Empty definition cannot be changed.
3. When you have angular components with OnPush (which is faster), the component is only rerendered when the input value is changed to a new instance. Therefore you cannot clear the data and update it. You have to construct a new object instead.

pjmagee:
> I am not sure if you have seen my comments, but these 3 things are important:
> 
> 1. Because so many components need this feature it is important to extract the methods to a separate file.
Yes, I noticed this and this will be a refactoring task as part of this, probably I will place this export function in the language state file.
> 2. The Empty definition cannot be changed.
If you are referring to the IsValid then I may need to make a new export function, i was getting strange behaviour with not checking if an array length was 0 if the field type is an array. If you dont mean this, than I am not sure. 
> 3. When you have angular components with OnPush (which is faster), the component is only rerendered when the input value is changed to a new instance. Therefore you cannot clear the data and update it. You have to construct a new object instead.
I understand the concept of OnPush and change detection, but did not think about the issue you just mentioned, which actually now makes sense to me. so I will create a new Map as you said, so the trigger will occur, instead of clearing.

SebastianStehle:
> > The Empty definition cannot be changed.
>
> If you are referring to the IsValid then I may need to make a new export function, i was getting strange behavior with not checking if an array length was 0 if the field type is an array. If you dont mean this, than I am not sure.

At the top right corner of each field there is a button to clear the field. This option is enabled, when the field value is defined, aka "IsValid". A field value is defined when it is not `null` and not `undefined`.

For some fields it is hard to distinguish it visually. For example an input field looks the same, whether the string is empty, has only whitespaces, is `null` or is `undefined`. But for the API it makes a big difference. For each language you can define a fallback language or multiple fallback languages. If a field is not defined for one language the result is replaced with the value from the fallback language.

For example: You have a schema for a page with an localized array field for advertisement:

 * If you set the field to empty (`[ ]`) for German then you will not see any advertisement for this language.
 * If you set the field to `null`, then the default handling for languages will replace the field with the English advertisement when queried by the API.

Therefore it is a big difference and we no do treat empty arrays as empty.

In your first implementation and with the current implementation you have changed the definition of a "defined value". Now every empty array is also undefined, which is not consistent with the API anymore. Therefore you cannot change the "isValid" value and we should also have the same behavior for the language selector and the clear-button.

You could argue that we should highlight the difference between empty values and undefined values, but this is another topic or task.

>> Because so many components need this feature it is important to extract the methods to a separate file.
> 
> Yes, I noticed this and this will be a refactoring task as part of this, probably I will place this export function in the language state file.

I would say it is probably more related to content. You can also just make a helper file somewhere in the content feature area.

pjmagee:
> > > The Empty definition cannot be changed.
> > 
> > 
> > If you are referring to the IsValid then I may need to make a new export function, i was getting strange behavior with not checking if an array length was 0 if the field type is an array. If you dont mean this, than I am not sure.
> 
> At the top right corner of each field there is a button to clear the field. This option is enabled, when the field value is defined, aka "IsValid". A field value is defined when it is not `null` and not `undefined`.
> 
> For some fields it is hard to distinguish it visually. For example an input field looks the same, whether the string is empty, has only whitespaces, is `null` or is `undefined`. But for the API it makes a big difference. For each language you can define a fallback language or multiple fallback languages. If a field is not defined for one language the result is replaced with the value from the fallback language.
> 
> For example: You have a schema for a page with an localized array field for advertisement:
> 
> * If you set the field to empty (`[ ]`) for German then you will not see any advertisement for this language.
> * If you set the field to `null`, then the default handling for languages will replace the field with the English advertisement when queried by the API.
> 
> Therefore it is a big difference and we no do treat empty arrays as empty.
> 
> In your first implementation and with the current implementation you have changed the definition of a "defined value". Now every empty array is also undefined, which is not consistent with the API anymore. Therefore you cannot change the "isValid" value and we should also have the same behavior for the language selector and the clear-button.
> 
> You could argue that we should highlight the difference between empty values and undefined values, but this is another topic or task.
> 
> > > Because so many components need this feature it is important to extract the methods to a separate file.
> > 
> > 
> > Yes, I noticed this and this will be a refactoring task as part of this, probably I will place this export function in the language state file.
> 
> I would say it is probably more related to content. You can also just make a helper file somewhere in the content feature area.

Thank you for the feedback, i will pick up more of this tomorrow :)

pjmagee:
> > > The Empty definition cannot be changed.
> > 
> > 
> > If you are referring to the IsValid then I may need to make a new export function, i was getting strange behavior with not checking if an array length was 0 if the field type is an array. If you dont mean this, than I am not sure.
> 
> At the top right corner of each field there is a button to clear the field. This option is enabled, when the field value is defined, aka "IsValid". A field value is defined when it is not `null` and not `undefined`.
> 
> For some fields it is hard to distinguish it visually. For example an input field looks the same, whether the string is empty, has only whitespaces, is `null` or is `undefined`. But for the API it makes a big difference. For each language you can define a fallback language or multiple fallback languages. If a field is not defined for one language the result is replaced with the value from the fallback language.
> 
> For example: You have a schema for a page with an localized array field for advertisement:
> 
> * If you set the field to empty (`[ ]`) for German then you will not see any advertisement for this language.
> * If you set the field to `null`, then the default handling for languages will replace the field with the English advertisement when queried by the API.
> 
> Therefore it is a big difference and we no do treat empty arrays as empty.
> 
> In your first implementation and with the current implementation you have changed the definition of a "defined value". Now every empty array is also undefined, which is not consistent with the API anymore. Therefore you cannot change the "isValid" value and we should also have the same behavior for the language selector and the clear-button.
> 
> You could argue that we should highlight the difference between empty values and undefined values, but this is another topic or task.
> 
> > > Because so many components need this feature it is important to extract the methods to a separate file.
> > 
> > 
> > Yes, I noticed this and this will be a refactoring task as part of this, probably I will place this export function in the language state file.
> 
> I would say it is probably more related to content. You can also just make a helper file somewhere in the content feature area.

Regarding the IsValid, the only way im getting visual benefit from a UX perspective is bolding languages where there is actual content set for that language. So if something localised is a collection, and there is an empty set, then there is no localised content for it. If the array does have 1 item, i know its localized and can bolden the language in the dropdown. 

If we go back to only using the `IsValid` or now changed to `IsValidFormValue` for also the language dropdown, are you expecting the language dropdown to still give a good UX? I am sure when i did not include this array check, i was not getting results I had expected and things were not bold when they should be, or maybe I am just getting confused.

SebastianStehle:
Closed due to inactivity

PR #784, opened 2021-11-01
Nested Schema Categories for new Schema State

bacon1986:
This now works with a lot less code and without the emit.
image

Filtering updates the schema counts as expected:
image

~~It still suffers from the long schema name issue but I'll look into that now.~~ Fixed
image

Thanks

### DIFF
--- a/frontend/app/shared/components/schema-category.component.ts
+++ b/frontend/app/shared/components/schema-category.component.ts
@@ -22,7 +22,7 @@ const ITEM_HEIGHT = 2.5;
 })
 export class SchemaCategoryComponent implements OnChanges {
     @Output()
-    public remove = new EventEmitter();
+    public remove = new EventEmitter<string>();
 
     @Input()
     public schemaCategory: SchemaCategory;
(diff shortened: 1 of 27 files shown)


frontend/app/shared/components/schema-category.component.ts
```diff
@@ -49,13 +51,19 @@ export class SchemaCategoryComponent implements OnChanges {
...
     public ngOnChanges() {
+        this.visibleCount = this.getCount(this.schemaCategory);
         if (this.schemaCategory.schemas.length < this.schemaCategory.schemaTotalCount) {
             this.isCollapsed = false;
         } else {
             this.isCollapsed = this.localStore.getBoolean(this.configKey());
         }
     }
 
+    private getCount(category: SchemaCategory): number {
```
bacon1986, line 62: This method replaced the previous `emit` which allowed the controls to update either other when they filtered, but now the filtering is all in the state that's not needed any more 👍

bacon1986:
I've just noticed that adding multi-level categories at the bottom of the schema list doesn't work so I'll look into that.

pjmagee:
Okay, this is really nice actually.

SebastianStehle:
I have merged it and made a few improvements to also calculate the filter count outside of the components.

PR #785, opened 2021-11-02
Reference item schema label hover minor enhancement

pjmagee:
Default view
image

Before with title attribute
image

After, with hover and expand. We found some users wanted to copy schema names, this allows it.
image
image


SebastianStehle:
Thank you. I also added a little bit of transition and removed the title.

PR #808, opened 2021-12-09
Updating MongoDB.Driver and MongoDB.Driver.GridFS to newer patch version

mtareq2:
Updating MongoDB.Driver and MongoDB.Driver.GridFS from 2.13.2 to the patch release version 2.13.3


SebastianStehle:
Thank you very much. Is there a bugfix you are looking for? Just curious, going to merge it anyway.

mtareq2:
Thanks for merging.
I was facing a weird issue, Squidex was failing to start because it wasn't able to find assembly `MongoDB.Driver, Version=2.13.3.0` I checked all mongo dlls in the projects and all of them were version `2.13.2.0` and couldn't find any dependency with version `2.13.3.0` 
I checked the releases here https://mongodb.github.io/mongo-csharp-driver/ and found that `2.13.2.0` isn't listed as a release, so I thought it's a good thing anyway to update it, but I still don't know why it was looking for the wrong version!
Nothing special about `web.config` but it worth mentioning that  in my case squidex is hosted in IIS (I wasn't running squidex with the `dotnet` command).

SebastianStehle:
Weird, every build will run Squidex as part of the API tests.

PR #870, opened 2022-04-21
RBAC updates to support aggregated role bindings

jrlost:
RBAC updates to support aggregated role bindings when using the OidcRoleMapping with an external identity provider.

See https://support.squidex.io/t/support-multiple-roles-in-oidc-role-mapping/4166 for more details regarding this change.

### DIFF
--- a/backend/src/Squidex/Config/Authentication/OidcHandler.cs
+++ b/backend/src/Squidex/Config/Authentication/OidcHandler.cs
@@ -24,16 +24,18 @@ namespace Squidex.Config.Authentication
         {
             var identity = (ClaimsIdentity)context.Principal!.Identity!;
 
-            if (!string.IsNullOrWhiteSpace(options.OidcRoleClaimType) && options.OidcRoleMapping?.Count >= 0)
-            {
-                var role = identity.FindFirst(x => x.Type == options.OidcRoleClaimType)?.Value;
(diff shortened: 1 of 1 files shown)


backend/src/Squidex/Config/Authentication/OidcHandler.cs
```diff
@@ -26,14 +26,18 @@ public override Task TokenValidated(TokenValidatedContext context)
 
             if (!string.IsNullOrWhiteSpace(options.OidcRoleClaimType) && options.OidcRoleMapping?.Count >= 0)
             {
-                var role = identity.FindFirst(x => x.Type == options.OidcRoleClaimType)?.Value;
+                var roles = identity
+                    .FindAll(c => c.Type == options.OidcRoleClaimType)
```
SebastianStehle, line 30: Just a very small change request:

This can be changed to FindAll(options.OidcRoleClaimType)

SebastianStehle: or even

options.OidcRoleMapping.Where(r => identity.HasClaim(options.OidcRoleClaimType, r.Key))

jrlost: good catch, thanks

PR #896, opened 2022-07-08
Schema name validation message no longer says spaces are allowed

slalFe:
- Couple of 'pager' transalations automatically added to frontend_zh when running translate.bat


slalFe:
I am less confident with the Chinese translation as I just have to trust Google Translate.

PR #914, opened 2022-08-25
Clarify Folder Name Hint

zicklag:
The previous wording indicated that folder names must not be unique but what is actually meant is that folder names might not be unique.


SebastianStehle:
Hi, please run the translator to update all other files: https://docs.squidex.io/01-getting-started/contributing-and-developing/translating

zicklag:
There you go!

PR #941, opened 2022-11-17
Kubernetes Security Hardening

Strayfe:
We need to ensure that we offer users the ability to harden their security footprint when deploying this application to Kubernetes clusters and the like.

This pull request includes the changes that will offer support for dev/ops the ability to configure their installations via helm to include the security best practices outlined by the Kubernetes documentation.

https://kubernetes.io/docs/concepts/security/
https://kubesec.io/basics/

I have also corrected an environment variable that was preventing the application from starting up in a cluster and updated the chart repository for MongoDB following recent updates to helm.

Reference: https://helm.sh/blog/new-location-stable-incubator-charts/


backend/src/Squidex/Program.cs
```diff
@@ -56,6 +56,10 @@ public static void Main(string[] args)
 
                         serverOptions.ListenAnyIP(5000);
                     }
+                    else
```
Strayfe, line 59: Likely this will want to be expanded upon, just including in the initial changes to demonstrate how the application _needs_ to run on a non-privileged port (1-1024) as this requires root access.

backend/src/Squidex/Program.cs
```diff
@@ -56,6 +56,10 @@ public static void Main(string[] args)
...
+                    {
+                        serverOptions.ListenAnyIP(10080);
```
Strayfe, line 61: Likely this will want to be expanded upon, just including in the initial changes to demonstrate how the application needs to run on a non-privileged port (1-1024) as this requires root access.

helm/squidex/values.yaml
```diff
@@ -46,6 +46,12 @@ affinity: { }
 ## @skip clusterSuffix
 clusterSuffix: cluster.local
 
+## @skip podSecurityContext
+podSecurityContext: { }
```
Strayfe, line 50: An example of a **Pod Security Context** that follows the rules defined by the strictest policy outlined by Kubernetes **Restricted**:

```
podSecurityContext:
  seccompProfile:
    type: RuntimeDefault
  runAsNonRoot: true
  runAsUser: 10000
  runAsGroup: 10000
  fsGroup: 10000
```

helm/squidex/values.yaml
```diff
@@ -46,6 +46,12 @@ affinity: { }
...
+
+## @skip containerSecurityContext
+containerSecurityContext: { }
```
Strayfe, line 53: An example of a **Container Security Context** that follows the rules defined by the strictest policy outlined by Kubernetes (**Restricted**):

```
containerSecurityContext:
  capabilities:
    drop:
      - ALL
    add:
      - NET_BIND_SERVICE
  allowPrivilegeEscalation: false
```

Strayfe:
This PR is incomplete and should not yet be merged.
I need to check how we would go about configuring the mongodb-replicaset to also include these changes.

backend/src/Squidex/Program.cs
SebastianStehle: This would be a breaking change, but you can control it with environment variables: https://andrewlock.net/5-ways-to-set-the-urls-for-an-aspnetcore-app/

Strayfe: Yes, I will update this soon, I just wanted to get something up for you to see so we can discuss.

Strayfe: Updated to apply from configuration.

helm/squidex7/templates/deployment-worker.yaml
```diff
@@ -21,13 +21,17 @@ spec:
...
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          securityContext:
+            {{- toYaml .Values.containerSecurityContext | nindent 12 }}
           ports:
             - name: http
-              containerPort: 80
+              containerPort: 10080
```
SebastianStehle, line 34: This would be a breaking change, right? So we need a way to configure this. The setting should then also be mapped to ASPNETCORE__URLS environment variable.

helm/squidex7/values.yaml
```diff
@@ -224,6 +233,12 @@ mongodb-replicaset:
   ##
   replicas: 3
   
+  ## @skip podSecurityContext
+  podSecurityContext: { }
```
SebastianStehle, line 237: Should this not be described?

Strayfe: Yes, I will add annotations to the yaml soon.

Strayfe: Described.

helm/squidex7/templates/deployment-worker.yaml
Strayfe: I just reviewed the documentation for other deployment methods for this software and found that a few supporting changes may be required on the https://github.com/Squidex/squidex-hosting repository to update the docker-compose files to proxy 80 and 443 to their respective ports exposed by the Dockerfile.

For example, I would propose changing the [docker-compose.yaml](https://github.com/Squidex/squidex-hosting/blob/master/docker-compose/docker-compose.yml) from:
```
  squidex_proxy:
    image: squidex/caddy-proxy
    ports:
      - "80:80"
      - "443:443"
```

to:
```
  squidex_proxy:
    image: squidex/caddy-proxy
    ports:
      - "80:10080"
      - "443:10443"
```

Strayfe: Again, this would be a result of the changes to the Dockerfile, not these helm values.

Strayfe:
It may be worthwhile to note that the [squidex documentation](https://docs.squidex.io/01-getting-started/installation/platforms/install-on-kubernetes) states that installation on Kubernetes supports version 1.14+ but the changes in this PR would only apply to Kubernetes to 1.21+ due to the deprecation (and removal) of the previous [Pod Security Policy](https://kubernetes.io/docs/concepts/security/pod-security-policy/) feature which the [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) replaces.

SebastianStehle:
I guess you could use some conditional expression in the template to solve that.

Strayfe:
Aye, that is true, I have recently seen someone using a conditional based on the detected version of k8s, I will add this condition to the templates.

SebastianStehle:
Will you provide a PR for the Mongo Helm chart?

Strayfe:
I was just looking at what would be required to do this work and I was about to suggest upgrading to the bitnami variant as it includes the correct `securityContext` injection.

Would you prefer I attempt to upgrade the MongoDB helm charts as a separate PR to this?

SebastianStehle:
> Would you prefer I attempt to upgrade the MongoDB helm charts as a separate PR to this?

Not, it is up to you.

Strayfe:
Well, upgrading the mongodb chart is proving to be a nightmare :)

SebastianStehle:
Yes, I already spend half an hour with that and then realized that I do not want to waste my time with that.

Strayfe:
Turns out my nightmares were nothing to do with the MongoDB helm charts at all, I was experiencing strange issues caused by my local `btrfs` file-system and a known bug in docker 😞 

So, all the changes I have made now yield the desired results for me.

If you deploy these helm charts with values overrides of the following: 

```
podSecurityContext:
  seccompProfile:
    type: RuntimeDefault
  runAsNonRoot: true
  runAsUser: 10000
  runAsGroup: 10000
  fsGroup: 10000
containerSecurityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
    add:
      - NET_BIND_SERVICE

mongodb:
  podSecurityContext:
    enabled: true
    fsGroup: 1001
    sysctls: [ ]
    seccompProfile:
      type: RuntimeDefault
    runAsNonRoot: true
    runAsUser: 1001
  containerSecurityContext:
    enabled: true
    runAsUser: 1001
    runAsNonRoot: true
    allowPrivilegeEscalation: false
    capabilities:
      drop:
        - ALL
      add:
        - NET_BIND_SERVICE
    
  arbiter:
    podSecurityContext:
      enabled: true
      fsGroup: 1001
      sysctls: [ ]
      seccompProfile:
        type: RuntimeDefault
      runAsNonRoot: true
      runAsUser: 1001
    containerSecurityContext:
      enabled: true
      runAsUser: 1001
      runAsNonRoot: true
      allowPrivilegeEscalation: false
      capabilities:
        drop:
          - ALL
        add:
          - NET_BIND_SERVICE
```

and check the output of 
```
kubectl label --dry-run=server --overwrite ns squidex pod-security.kubernetes.io/enforce=restricted
```

we can see that the namespace can be labelled successfully without errors or warnings
image

SebastianStehle:
Great. So you think it is ready to be merged?

Strayfe:
Yeah, I think it's ready.

Remember the `Dockerfile` now exposes 10080 and 10443 instead of 80 and 443 so other deployment methods like `docker compose` will need their configurations updating to map external 80 & 443 to internal 10080 and 10443 respectively.

SebastianStehle:
> Remember the Dockerfile now exposes 10080 and 10443 instead of 80 and 443 so other deployment methods like docker compose will need their configurations updating to map external 80 & 443 to internal 10080 and 10443 respectively.

You can add additional ports, but I do not want a breaking change. This would require an 8.0 version, just for two ports.

Strayfe:
Port 80 and 443 are "privileged ports" which require root access which is what this PR is trying to correct as containers running as root are a security concern. We cannot add them as additional ports, we need to make sure we are not using ports in the 1-1024 range.

However, it is my understanding that the `NET_BIND_SERVICE` capability can allow a non-root user the ability to bind to privileged ports so maybe I can do some testing to see if we can avoid needing to change the port, for now.

SebastianStehle:
A change is not an option. If you really have to change that, then you have to change the CI to publish a second container with other ports.

Strayfe:
I have managed to get it working without needing to change the ports.

SebastianStehle:
But it means that you need at least Squidex 7.5 to run this, right? Because otherwise the port mapping would not work.

helm/squidex/templates/_helpers.tpl
```diff
@@ -26,8 +26,8 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- define "squidex.mongoDefaultHostname" }}
   {{- $release := .Release}}
   {{- $clusterSuffix := .Values.clusterSuffix}}
-  {{- range $index, $i := until (int (index .Values "mongodb-replicaset").replicas) }}
-    {{- $replica := printf "%s-mongodb-replicaset-%d.%s-mongodb-replicaset.%s.svc.%s" $release.Name $i $release.Name $release.Namespace $clusterSuffix }}
+  {{- range $index, $i := until (int (index .Values "mongodb").replicaCount) }}
+    {{- $replica := printf "%s-mongodb-%d.%s-mongodb-headless.%s.svc.%s" $release.Name $i $release.Name $release.Namespace $clusterSuffix }}
```
SebastianStehle, line 30: Why is this needed? Is this not a breaking change?

Dockerfile
```diff
@@ -87,10 +87,28 @@ WORKDIR /app
...
+# it is the dotnet binary that requires the capability to launch a kestrel server on privileged ports
+RUN setcap CAP_NET_BIND_SERVICE=+ep /usr/share/dotnet/dotnet
+
+# expose privileged ports
 EXPOSE 80
 EXPOSE 443
 EXPOSE 11111
 
+# add a group and a user > 10000 to avoid conflicts with system accounts
+RUN groupadd --gid 10000 --system runner
```
SebastianStehle, line 102: Why is something like this not part of the base image from Microsoft?

Strayfe:
> But it means that you need at least Squidex 7.5 to run this, right? Because otherwise the port mapping would not work.

The port mappings have not changed, I don't understand what you mean. You asked me to ensure the ports were not changed as it would be regarded as a breaking change?

SebastianStehle:
I just don't understand what these two lines do in the Dockerfile.

```
# set capability of NET_BIND_SERVICE to allow containers to bind and expose privileged ports
# set net_bind both [e]ffective & [p]ermitted
RUN apt install libcap2 libcap2-bin -y
# it is the dotnet binary that requires the capability to launch a kestrel server on privileged ports
RUN setcap CAP_NET_BIND_SERVICE=+ep /usr/share/dotnet/dotnet
```

Dockerfile
Strayfe: This is something Microsoft seem to be aware of and they have been very recently (Aug 22) making new images that are more secure by default like the recent `Ubuntu-Chiseled` containers but sadly these non-root changes seem to have only been made to Ubuntu and the likes of Alpine and Debian have currently been ignored.

https://devblogs.microsoft.com/dotnet/dotnet-6-is-now-in-ubuntu-2204/#non-root-images
https://github.com/dotnet/dotnet-docker/blob/fd39ea3eece8c8653348a717446ac41e93633784/src/runtime-deps/6.0/jammy-chiseled/arm64v8/Dockerfile#L13-L26

My guess would be that they are just slightly behind on container security best practices and will slowly adopt them into their offerings.

helm/squidex/templates/_helpers.tpl
Strayfe: The schema for the bitnami mongodb charts is different to the version that was deployed previously.

The change would require consumers to update their `values.yaml` files if they have made custom changes to their deployments, yes.

The new service that is deployed using the bitnami charts is appended with `-headless` and so requires an update to the connection string so that it can be addressed by DNS internally correctly. 
image

The other changes are based on how helm uses naming conventions to apply value substitution to child charts from the parent configuration, because the schema has changed on the child, we need to apply correct values.

Strayfe:
> I just don't understand what these two lines do in the Dockerfile.

probably self-explanatory, this installs libcap2-bin which contains cmdlets called `setcap` and `getcap` which allows us to set certain linux capabilities for non-root users.
```
RUN apt install libcap2 libcap2-bin -y
```

Typically, when you remove all capabilities from a non-root user, they cannot perform root actions on a linux machine without running sudo or changing to the root user, like binding privileged ports such as 1-1024.

The following command `setcap` applies the root capability to bind privileged ports to the non-root user.
The capability is `CAP_NET_BIND_SERVICE`.
The specific binary that I am specifying has the ability to bind privileged ports is the dotnet runtime at `/usr/share/dotnet/dotnet`.

```
RUN setcap CAP_NET_BIND_SERVICE=+ep /usr/share/dotnet/dotnet
```

If you do not take these steps to allow a service like dotnet to bind privileged ports on a linux container after removing root capabilities and running as a non-root user, you will find that dotnet will not run and throw an exception stating that `Permission denied` when trying to bind the port, e.g. 80.

SebastianStehle:
Okay, got it.

Strayfe:
The changes I have made apply to both helm charts `squidex` and `squidex7`, the single dockerfile is referenced by both installations and I have updated both chart templates, unless I have missed something?

SebastianStehle:
Sorry, it was more a question ;)

Strayfe:
Ouh 😛 

For anyone who wants to deploy these changes, they would simply need to grab the correct helm chart from this repo and apply their own custom values over the top, whether that is `squidex` or `squidex7`.

helm/squidex/templates/_helpers.tpl
SebastianStehle: But then you would loose your previous data if you make an update, right? I am not sure what the convention for helm versioning is in this case.

Strayfe:
Oh I might have had a small oversight!

We will need a new docker image publishing to https://hub.docker.com/r/squidex/squidex/ and then we need to reference that new image in the helm charts `values.yaml`.

SebastianStehle:
Yeah, but only if the users override some of these security settings. So basically on older versions it would have no impact. Perhaps there are some setting which could be set?

Dockerfile
SebastianStehle: To be honest I have never heard about it. It does not mean anything because I work in small startups where every developer is superadmin anyway ;) ... but even from my large customers with 100+ developers I have no heard about this before.

Strayfe: I am not sure if it is just a blind-spot in the industry or if the concept is immature but it certainly needs more visibility.

Containers are just small linux instances capable of doing almost anything a regular linux machine is capable of, so when we allow containers to run as root we allow attackers a lot of room to escalate their attacks into the cluster and potentially even further.

For example, should an attacker find a vulnerability in an application that allows them to execute a shell in a container that has root access, they could use `apt` to install `curl`, `git`, `kubctl`, etc. They could then download malicious payloads into the container, enumerate other containers/pods in the cluster and widen their attack internally to find your persistent data stores where your sensitive data is stored, try to take down parts of the cluster, etc.

Sure, whilst such an attack would require another vulnerability to exist allowing the attacker to execute commands inside the container, we should never assume that this is not possible. The principal of "security by design" is to always expect attacks and an attacker is not going to stop once they have a foothold in a system, they will try to escalate that attack as far and wide as possible. We need to ensure that every layer of the onion is secure by design, and container security is an additional layer that needs to be secured.

An interesting read that has parallels to this topic:
https://www.cyberark.com/resources/threat-research-blog/how-i-hacked-play-with-docker-and-remotely-ran-code-on-the-host

The changes I am making in this PR, whilst mainly aimed at Kubernetes clusters and helm charts, are facilitated by the changes to the container image `Dockerfile` and with the appropriate supplementary changes, could be used to secure other deployment methods like `docker compose`, `docker swarm`, etc.

helm/squidex/templates/_helpers.tpl
Strayfe: Helm versioning would really only apply here if you had a helm repository I think?

But yes, I think users wanting to deploy these changes to their cluster would need to potentially backup their data (they should be doing this anyway!!) and restore after applying this change.

I might have a look in the morning to see if there would be a way to keep the name of the `Persistent Volumes` and `Persistent Volume Claims` the same between these deployments (if they are not already), might take a little while as I will need to deploy `Squidex7`, add data to the CMS and then try and upgrade to this branch and see the outcome.

I will get back to you on this point and move the PR back into "Draft".

SebastianStehle:
Hi @Strayfe ,

what is the status?

Strayfe:
Afternoon Seb-asaurus,

I managed to get the `Persistent Volume Claims` and `Persistent Volumes` to use the same naming convention which means that they could be re-bound to the new pods that get spun up, I also managed to mount the volume into the container under the same paths `/data/db`, however, the differences in the mongodb charts has stumped me and I cannot seem to get the bitnami version to read the data from the original.

Right now it looks like the only way to approach this upgrade would be to backup & restore which I believe you will say is a breaking change.

I have run out of time to look at this currently so I am either just going to run this fork in our environment or I might have to look at a solution that already has the security fixes in place as we cannot risk container escape in our clusters.

You can do what you want with this PR, if you decide to make a v8 then maybe we could re-approach it.

SebastianStehle:
Hi,
we could also deploy it with a new tag like squidex-secure or so. Then we would avoid all compatibility issues in the helm chart.

PR #964, opened 2023-01-09
Update permissions on external-callback login

slalFe:
- Previously users that already existed would not have their permissions updated in Squidex

Hello, this has not been tested! Just wanted to know if this makes sense or not? And if it does and works...well feel free to merge!

This is in relation to: https://support.squidex.io/t/openiddict-change-permissions-claims-for-existing-users/4295

Is there test coverage for this class? If so I could not find it, would have preferred to have unit tests covering the scenarios so I could know for sure `Equals` definitely works for Claims as I hope! Assume unit testing around things like `SignInManager` can get a bit complicated?

### DIFF
--- a/backend/src/Squidex/Areas/IdentityServer/Controllers/Account/AccountController.cs
+++ b/backend/src/Squidex/Areas/IdentityServer/Controllers/Account/AccountController.cs
@@ -241,19 +241,14 @@ public sealed class AccountController : IdentityServerController
         {
             user = await userService.FindByLoginAsync(login.LoginProvider, login.ProviderKey, HttpContext.RequestAborted);
 
-            if (user != null)
+            if (user != null && identityOptions.OidcOverridePermissionsWithCustomClaimsOnLogin)
             {
-                var squidexClaimsFromLoginPrincipal = login.Principal.Claims.GetSquidexClaims();
-
-                if (user.Claims.GetSquidexClaims().Equals(squidexClaimsFromLoginPrincipal))
+                var values = new UserValues
                 {
-                    var values = new UserValues
-                    {
-                        CustomClaims = squidexClaimsFromLoginPrincipal.ToList()
-                    };
+                    CustomClaims = login.Principal.Claims.GetSquidexClaims().ToList()
+                };
 
-                    user = await userService.UpdateAsync(user.Id, values, false, HttpContext.RequestAborted);
-                }
+                user = await userService.UpdateAsync(user.Id, values, false, HttpContext.RequestAborted);
             }
         }
         else
--- a/backend/src/Squidex/Areas/IdentityServer/Controllers/Account/AccountController.cs
+++ b/backend/src/Squidex/Areas/IdentityServer/Controllers/Account/AccountController.cs
@@ -240,6 +240,21 @@ public sealed class AccountController : IdentityServerController
         if (isLoggedIn)
         {
             user = await userService.FindByLoginAsync(login.LoginProvider, login.ProviderKey, HttpContext.RequestAborted);
+
+            if (user != null)
+            {
+                var squidexClaimsFromLoginPrincipal = login.Principal.Claims.GetSquidexClaims();
+
+                if (user.Claims.GetSquidexClaims().Equals(squidexClaimsFromLoginPrincipal))
+                {
+                    var values = new UserValues
+                    {
+                        CustomClaims = squidexClaimsFromLoginPrincipal.ToList()
+                    };
+
+                    user = await userService.UpdateAsync(user.Id, values, false, HttpContext.RequestAborted);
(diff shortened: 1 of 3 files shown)


backend/src/Squidex/Areas/IdentityServer/Controllers/Account/AccountController.cs
```diff
@@ -240,6 +240,21 @@ public async Task<IActionResult> ExternalCallback(string? returnUrl = null)
         if (isLoggedIn)
         {
             user = await userService.FindByLoginAsync(login.LoginProvider, login.ProviderKey, HttpContext.RequestAborted);
+
+            if (user != null)
+            {
+                var squidexClaimsFromLoginPrincipal = login.Principal.Claims.GetSquidexClaims();
+
+                if (user.Claims.GetSquidexClaims().Equals(squidexClaimsFromLoginPrincipal))
```
SebastianStehle, line 248: I think Equals does not work here. Must be something like SetEquals or so. It would also override all claims that have been set manually, not sure if this is intended.

slalFe: Ah good point, for our use case we absolutely want permissions in Squidex to exactly match what is in our providers configuration but perhaps others may not want the same behaviour; could break existing user's setups if they have only manually setup permissions and just use the external provider to login. Does this mean it'd need a new configuration option called something like `oidcSyncPermissionsWithClaimsOnLogin` that's false by default?

Thanks for the tip around `SetEquals`, knew I was optimistic just putting `Equals` in there! Will try and actually test it this time before I push it up for review.

SebastianStehle: Yes, another option would be good.

`GetSquidexClaims` does return `IEnumerable<Claim>` which does not implement Equals, therefore you have to find another method for that. Also better call ToList() if you make the call twice to the same user.

slalFe: Added configuration option but ran out of time today (and not working tomorrow) so simply removed my invalid claims comparison for now to see what you think about relying on your existing `SyncClaims` code to only update what is needed which is tried and tested? Seems like it should be pretty quick to go through `UpdateAsync` and then `SyncClaims`, and only happens on login for people making use of this new setting, however I can understand how any performance overhead should be avoided if possible.

backend/src/Squidex/Config/MyIdentityOptions.cs
```diff
@@ -59,6 +59,8 @@ public sealed class MyIdentityOptions
 
     public bool OidcGetClaimsFromUserInfoEndpoint { get; set; }
 
+    public bool OidcOverridePermissionsWithCustomClaimsOnLogin { get; set; }
```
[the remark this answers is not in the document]
slalFe: Just wanted it to be as explicit as possible to remove any ambiguity!

One thing I am unsure of is whether this code change could potentially impact other providers, or is it completely specific to the OIDC login process? If it applies to others then perhaps the `oidc` prefix and location in configuration should change.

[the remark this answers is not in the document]
SebastianStehle: It is the only provider where you have control over the claims.

SebastianStehle:
Is it ready to be merged?

slalFe:
> Is it ready to be merged?

Sorry forgot to say I was going on holiday for a week! As such I still haven't tested it, and despite the removal of the invalid equality logic making it a lot simpler and hopefully safer change (unless there are performance issues) we should obviously take on the burden of initial testing. Will hopefully find time today although I have quite a few PRs and emails to catch up on!

slalFe:
There will be a slight delay on that testing as our base images don't seem to support .NET 7 yet so we are stuck on Squidex v7.2...would you be happy with testing just being done on a 7.2 patch?

SebastianStehle:
I have no plans to test that. If you test that and it looks okay, I am happy to merge it.

slalFe:
Just an update; our base images do support .NET 7 but our Node version was only on 14.16.0 and apparently Squidex now needs at least 14.20.0. Is there a version you specifically target? We are now testing the build with latest LTS version 18.13.0 and so far it is green...

slalFe:
Ended up targeting Node v16.15.0 as v18.13.0 gave a compilation error and we had another asset already using that older version so was readily available and it built OK.

More importantly I could then actually test this code change and it worked(!); flag as false meant permissions are not changed on login (the pre-existing behaviour), flag as true caused permissions to be overwritten with the incoming claims.

Do you want this branch rebased on latest master before merging assuming you are happy with my testing?

PR #973, opened 2023-02-14
Added custom ID re-use test

slalFe:
- Specifically this is when providing a custom ID that has previously been used by content that has since been permanently deleted
- Added test cases for custom ID as our issue may be caused by use of special characters


slalFe:
Couldn't actually run it sorry! In a meeting atm but will have a look at running the tests myself tomorrow (any pointers would be greatly appreciated!) as it's obviously quite an important bit of coverage we should bear in mind when contributing.

SebastianStehle:
I think it is solved now.

PR #1041, opened 2023-11-14
Add option to run as nonroot

cfnarmiek:
I would like to be able to run Squidex as nonroot user.

MongoDB is already running as nonroot. I have seen [this](https://github.com/Squidex/squidex/pull/941) previous attempt and would like to propose this change which allows running nonroot without breaking existing functionality.


SebastianStehle:
Thanks a lot.

PR #1052, opened 2023-12-07
Set secure policy on antiforgery cookie to always

MuchMilk:
One-liner to set the cookie secure policy to always on the antiforgery cookie.

### DIFF
--- a/backend/src/Squidex/Areas/IdentityServer/Config/IdentityServerServices.cs
+++ b/backend/src/Squidex/Areas/IdentityServer/Config/IdentityServerServices.cs
@@ -126,6 +126,8 @@ public static class IdentityServerServices
             var identityOptions = c.GetRequiredService<IOptions<MyIdentityOptions>>().Value;
 
             options.SuppressXFrameOptionsHeader = identityOptions.SuppressXFrameOptionsHeader;
+
+            options.Cookie.SecurePolicy = CookieSecurePolicy.Always;
         });
 
         services.Configure<OpenIddictServerOptions>((c, options) =>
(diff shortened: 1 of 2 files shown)


kjs73:
👍

MuchMilk:
Could also only set the cookie secure policy in prod?

SebastianStehle:
You could try to fix the tests for that (https might work) or you only set the header based on the protocol of the base URL.

backend/src/Squidex/Areas/IdentityServer/Config/IdentityServerServices.cs
```diff
@@ -127,7 +128,13 @@ public static void AddSquidexIdentityServer(this IServiceCollection services)
 
             options.SuppressXFrameOptionsHeader = identityOptions.SuppressXFrameOptionsHeader;
 
-            options.Cookie.SecurePolicy = CookieSecurePolicy.Always;
+            // Set antiforgery cookie secure policy to always for https
+            var urlsOptions = config.GetSection("urls").Get<UrlOptions>() ?? new ();
```
SebastianStehle, line 132: Don't use config here. Try to inject IOptions<UrlOptions> or `IUrlGenerator` (see in the same file), because there should be only one call where the options is binded to the config.

MuchMilk: Agreed its cleaner without config - see changes

SebastianStehle:
LGTM

MuchMilk:
@SebastianStehle nice, thankyou! Is there any kind of timeline for when the next release including this fix will be released?

SebastianStehle:
No, sorry...but you can just use the current dev version, it is going through a lot of automated tests

MuchMilk:
@SebastianStehle just saw the new release - thanks again for your help :)

PR #1096, opened 2024-05-29
Dev workflow: add building for ARM platforms

kdavis:
Related to https://github.com/Squidex/squidex/pull/1093

I missed the change for the dev workflow, so hopefully builds in https://hub.docker.com/r/squidex/squidex/tags will also be arm64 too

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -39,6 +39,7 @@ jobs:
           build-args: "SQUIDEX__RUNTIME__VERSION=7.0.0-dev-${{ env.BUILD_NUMBER }}"
           cache-from: type=gha
           cache-to: type=gha,mode=max
+          platforms: linux/amd64,linux/arm64
           tags: squidex-local
 
       - name: Test - Start Compose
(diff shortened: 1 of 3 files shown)


SebastianStehle:
I think I have to revert that. The build fails all the time now.

SebastianStehle:
Btw: Why do you build twice?

kdavis:
@SebastianStehle If you revert, you'll also need to revert https://github.com/Squidex/squidex/pull/1093 (as that will have a similar issue)

I built twice as this commit https://github.com/Squidex/squidex/pull/1096/commits/2642d0a3b1fff8995ce765dfce33b6ab31b00b5e specifies multiple build platforms whilst also having a `load: true` (which conflicts when trying to build). In the second build, I run the same params again with multiple build platforms, but without the `load: true`

In retrospect, it'd probably be easier to just have a matrix with the build platforms... 🤔

SebastianStehle:
I have solved it by adding the cleanup step: https://github.com/Squidex/squidex/blob/master/.github/workflows/release.yml#L119

Matrix build would also run the tests twice I guess, so your approach is probably better.

PR #1100, opened 2024-06-02
Push ARM64 image rather than keep it in cache

kdavis:
https://github.com/Squidex/squidex/actions/runs/9332715416/job/25688888768#step:21:2936 Seems the removal of the `load: true` means that the ARM64 build is kept inside the cache and not put into docker to publish.

> WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

This PR combines the last 3 steps & pushes at the end, so hopefully this will start pushing multi-platform. Following https://docs.docker.com/build/ci/github-actions/test-before-push/ this example here.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -140,15 +140,6 @@ jobs:
         run: |
           echo "BUILD_NUMBER=$(($BUILD_NUMBER + 6000))" >> $GITHUB_ENV
 
-      - name: Publish - Build for Multi-Platforms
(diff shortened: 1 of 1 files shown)


SebastianStehle:
If it works now can you also update Release.yml?

kdavis:
@SebastianStehle Have just tested and working like a charm!
Will do the same for the release shortly 🙌

PR #1130, opened 2024-10-07
Open links in new tab by holding meta key

saturn4er:
On MacOS `Ctrl + click` opens context menu, so it would be nice also to add `Cmd + click` hotkey
https://support.squidex.io/t/implemented-improvement-for-mass-content-editing/5410/3


cheshir:
Bump. 
Quite important feature for our company, because all the managers are on mac devices.

PR #1147, opened 2024-12-03
Changing helm chart to use AppVersion out of the box

aravol:
This fixes issues where fresh installations must explicitly set and maintain the image tag, explicitly set it null in values, or install an older version of Squidex (currently `7.0.2`) hidden behind an apparently valid AppVersion

### DIFF
--- a/helm/squidex7/README.md
+++ b/helm/squidex7/README.md
@@ -46,16 +46,16 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Global parameters
 
-| Name                      | Description                    | Value             |
-| ------------------------- | ------------------------------ | ----------------- |
-| `service.type`            | Kubernetes Service type        | `ClusterIP`       |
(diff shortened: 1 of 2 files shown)


SebastianStehle:
Thank you, but what is "apps Chart Version"?

aravol:
Per helm

```
appVersion: The version of the app that this contains (optional). Needn't be SemVer. Quotes recommended.
```

As a kube admin, when I install a helm chart, I expect this app version to correspond to the tag of the primary application container, _especially_ in  cases like Squidex7 where it has been updated with each release. However, because `values.yaml` contains a default `image.tag` of `7.0.2`, this is not the case, and we end up running an old Squidex unless explicitly setting our own `image.tag` to `null`

This change addresses this simply by removing that default value. Since the appVersion has been maintained on `Chart.yaml` and the templates already use `default`, this default is really the only problem

PR #1207, opened 2025-03-31
Feature/ability to disable OIDC "profile" scope

osipenkom:

PR #1208, opened 2025-04-01
Feature/ability to disable OIDC profile scope

osipenkom:
To prevent deletion of default scope "profile" we need to disable this setting `oidcDisableProfileScope: false` by default.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -37,7 +37,7 @@ jobs:
         uses: actions/checkout@v4.2.2
 
       - name: Prepare - Setup dotnet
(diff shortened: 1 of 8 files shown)


osipenkom:
@SebastianStehle Hello, during previos PR (https://github.com/Squidex/squidex/pull/1207) I have missed something and that will change default external OIDC behaviour: it will disable "profile" scope by default according to the setting `oidcDisableProfileScope": true,`. To prevent deletion of default scope "profile" we need to disable this setting by default and then everything will work as before. It must be `oidcDisableProfileScope": false,` by default. Please merge this 1 row change asap. Thank you!

PR #1329, opened 2026-08-11
Fix interactive login on SQL databases by widening the OpenIddict token type

RTJoe:
## Problem

Interactive logins fail on every SQL database. The user completes the external provider round trip and then lands on the error page:

> Operation failed. We are really sorry that something went wrong.

The request fails while OpenIddict stores the authorization code:

```
Microsoft.EntityFrameworkCore.DbUpdateException
  at OpenIddict.EntityFrameworkCore.OpenIddictEntityFrameworkCoreTokenStore`3.CreateAsync
  at OpenIddict.Server.OpenIddictServerHandlers.Protection.CreateTokenEntry.HandleAsync
  at OpenIddict.Server.OpenIddictServerHandlers.GenerateAuthorizationCode.HandleAsync
```

with this inner exception:

```
Msg 2628: String or binary data would be truncated in table 'OpenIddictTokens', column 'Type'.
Truncated value: 'urn:openiddict:params:oauth:token-type:authorizati'.
```

## Cause

`OpenIddict.AspNetCore` is referenced at `7.4.0`, but `Squidex.OpenIdDict.EntityFramework` is pinned at `7.2.1` and still declares `HasMaxLength(50)` for the token type. OpenIddict 7.4 persists the full token type identifiers rather than the short names it used before, and `urn:openiddict:params:oauth:token-type:authorization_code` needs 57 characters.

Official OpenIddict widened this column to 150 characters for the same reason, so the mismatch only exists because the fork has no 7.4 release.

This is visible in the stored data — the token types changed shape mid-flight, and no authorization code has been written since:

| Type | Count |
| --- | --- |
| `access_token` | old short name |
| `urn:ietf:params:oauth:token-type:access_token` | current, 45 chars, fits |
| `authorization_code` | old short name, last written before the upgrade |

Two things masked the severity:

- **The client credentials flow keeps working**, because the access token identifier is only 45 characters. Only interactive logins break.
- **MongoDB is unaffected**, because it does not enforce a length. This affects SQL Server, PostgreSQL and MySQL alike, which all cap the column at 50.

## Fix

- Override the token type length to 150 in `AppDbContext`, matching official OpenIddict. This lives in Squidex rather than waiting for a 7.4 release of the fork.
- Add the `WidenOpenIddictTokenType` migration for SQL Server, PostgreSQL and MySQL.
- Extend `EFOpenIddictTests` with a round trip over the real `TokenTypeIdentifiers.Private.AuthorizationCode` constant. The existing test passed because it used the legacy short `"authorization_code"` literal, so it never exercised the length.

The new test asserts the value read back from the database rather than only that `SaveChangesAsync` succeeded, because MySQL truncates silently outside strict mode, which would corrupt the token type instead of failing loudly.

## Verification

Reproduced and confirmed against a production SQL Server instance running `dev-8637`: inserting a token with the authorization code identifier fails with the error above, while an access token insert succeeds. After widening the column to 150 the authorization code insert succeeds and interactive login works again.

Note that the migration only repairs the schema. Existing deployments already broken by this need the migration to run, or the equivalent `ALTER TABLE`.

Automated tests were not run locally, as `Squidex.Data.Tests` requires Docker for its test containers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/src/Squidex.Data.EntityFramework/AppDbContext.cs
+++ b/backend/src/Squidex.Data.EntityFramework/AppDbContext.cs
@@ -7,6 +7,7 @@
 
 using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore;
+using OpenIddict.EntityFrameworkCore.Models;
 using Squidex.Assets.TusAdapter;
 using Squidex.Domain.Apps.Entities.Apps;
 using Squidex.Domain.Apps.Entities.Assets;
@@ -54,6 +55,7 @@ public abstract class AppDbContext(DbContextOptions options, IJsonSerializer jso
         builder.UseMigration();
         builder.UseNames(jsonSerializer, jsonColumnType);
         builder.UseOpenIddict();
+        builder.UseOpenIddictTokenType();
         builder.UseRequest(jsonSerializer, jsonColumnType);
         builder.UseRules(jsonSerializer, jsonColumnType);
         builder.UseSchema(jsonSerializer, jsonColumnType);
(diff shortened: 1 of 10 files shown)


kevinchalet:
To clarify, the maximum length of the `Type` column didn't change in 7.4.0 but in 7.0.0 (as a general rule, the schema is never updated in minor versions). This change was documented here: https://documentation.openiddict.com/guides/migration/60-to-70#add-and-apply-migrations-if-applicable.

